PR #24, opened 2019-07-02
Fix secret key name references in StatefulSet

annismckenzie:
This PR fixes the secret key name references for the PostgreSQL password as well as the autowizard JSON. One instance of the secret key reference was already using the correct name (see line 87 in the `statefulset.yaml` which I didn't touch).


annismckenzie:
Sorry about that – new to this whole Helm chart thing. It's been done, I hope just raising the patch level is sufficient?

PR #25, opened 2019-07-02
Add ability to use volume claim templates in Zammad StatefulSet

annismckenzie:
This allows one to use the StatefulSet's capability of specifying volume claim templates. This is a must for HA so that each pod gets its own storage. Right now, increasing the replica count in the StatefulSet results in unschedulable pods because the default claim uses `accessModes: ["ReadWriteOnce"]` so only the first pod can mount it successfully.

I know that this isn't the end of the story for HA. We might also need to test and document how this would actually work. I just wanted to get this in so it at least successfully schedules multiple replicas.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: zammad
-version: 1.1.2
+version: 1.1.3
 appVersion: 3.0.0
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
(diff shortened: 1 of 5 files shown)


monotek:
Zammad is not able to run with more than 1 replica as it would depend on shared storage.

Every pod would need to be able to read and write to the same volume so a persistent volume with access mode "ReadWriteMany" would be needed.

See also: https://github.com/zammad/zammad/issues/1601

So at the moment its imho pretty much useless to have multiple "ReadWriteOnce" volumes as every instance of Zammad would save a different state there leading to unpredictable behaviour.

annismckenzie:
Right, I'm a dummy – closing.

monotek:
Thanks for the pr anyway :-)

monotek:
fyi.... i've implemented volume claim templates now in: https://github.com/zammad/zammad-helm/commit/45256a96e2461c61c02b49f348ba26bf780e5fc5

PR #37, opened 2020-02-24
Properly render the storageClassName

pcfens:
Since `Values.persistence.storageClass` are wrapped in a `with` statement the scope changes, so setting `persistence.storageClass` causes an error when rendering the chart. Setting the value to `.` fixes that issue to match the new scope.


pcfens:
It appears that tests failed because I didn't bump the version number. If that needs to happen for testing I'm happy to, but it feels odd to assume that the version number will be bumped just because of this PR.

PR #55, opened 2020-09-11
Support providing imagePullSecrets for private registries

pcfens:

Support user provided imagePullSecrets, for when a custom image needs to be pulled from a private registry.

The standard convention seems to be that the `name:` component is included when someone provides a value, so I've done the same here rather than rendering it automatically.

#### Checklist
- [X] Chart Version bumped
- [X] Variables are documented in the README.md

### DIFF
--- a/.github/ct.yaml
+++ b/.github/ct.yaml
@@ -3,5 +3,5 @@ helm-extra-args: --timeout 1200s
 check-version-increment: true
 debug: true
 chart-repos:
-  - bitnami=https://charts.bitnami.com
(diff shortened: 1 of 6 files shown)


monotek (changes requested):
Please raise version to 2.3.0 as a new feature is added.

pcfens:
The failures don't seem directly related to the changes I've proposed - if I'm reading that incorrectly please let me know so I can get thing fixed.

monotek:
Yes, i've already restartet ci several times.
Need to have a deeper look later.

PR #58, opened 2020-09-16
remove hard-coded replica count and define antiAffinity for it

mskarbe:
The replica count of `zammad-*` pods was hardcoded to 1. The parameter `replicas` permits to set it without the need of patching the stateful set manually. 
Then, as well, it seemed to make sense to me to define default antiAffinity for it, so the pods would not be scheduled onto the same node.

  - fixes #

#### Checklist
- [ ] Chart Version bumped
- [ ] Variables are documented in the README.md

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: zammad
-version: 2.4.0
+version: 2.4.1
 appVersion: 3.4.0
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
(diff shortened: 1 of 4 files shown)


monotek:
How do you sync the files of the statefullset from one replica to the other?

mskarbe:
@monotek, thank you for your response!
Can you please give me some insights about what kind of files you are referring to? I was not aware that there was any particular sync required, the understanding that I was provided with was that those were ElasticSearch data pods, that could be scaled without additional steps.
So do I suppose now correctly that it is deliberately allowed to have only one replica of this sts? Can I ask why?

monotek:
Using several replicas of the containers would need them to have access to the same data but some data is stored in the filesystem, which is not shared between the containers.

This might give some more insight: https://github.com/zammad/zammad/issues/1601

mskarbe:
@monotek, now it is clear! 
Thank you very much for explaining and sorry for unfitting PR. :)

monotek:
It could work, if you use read write many storage calss, but as this is not so common we decided to better remove the option.

Maybe it makes sense to add it and add a warning, that you need special storage class to use it.

PR #61, opened 2020-09-25
Update railsserver container to use the correct helm variable

cabillman:

Use the same helm variable as described in values.yaml and other containers in the pod for the railsserver. We wanted to increase the concurrency and used the extraEnv but the setting was ignored.

#### Checklist
- [ ] Chart Version bumped
- [ ] Variables are documented in the README.md


monotek (changes requested):
Please raise chart version

monotek (approved):

PR #62, opened 2020-09-29
Enable scaling above 1 replica

pcfens:
In some cases there may be a desire to scale Zammad to more than one replica.

If more than one replica is configured sticky sessions should also be enabled. It's hard to test for this since different ingress controllers and load balancers handle sticky sessions in different ways, so I've documented that there are no warnings (hopefully folks read this portion)

Some ingress controllers require an annotation on the service itself (e.g. Traefik 2+), so I've added support for setting that too.

#### Checklist
- [X] Chart Version bumped
- [X] Variables are documented in the README.md

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: zammad
-version: 2.6.1
+version: 2.7.0
 appVersion: 3.5.0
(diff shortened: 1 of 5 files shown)


monotek:
Duplicate of: https://github.com/zammad/zammad-helm/pull/58

PR #70, opened 2020-11-11
Nginx headers

### LINKED ISSUE #69: Ability to add custom nginx configuration
Would you be willing to accept a patch that allows for inserting additional nginx configuration into the nginx configmap from helm values?

We run our instance of Zammad behind a Google IAP and would like the ability to rewrite some of the headers before it hits zammad.  I'm specifically targeting this area - https://github.com/zammad/zammad-helm/blob/master/zammad/templates/configmap-nginx.yaml#L52

I'm trying avoid the need to run an additional instance of nginx in front for the purpose of rewriting a few headers.

cabillman:
Allows users of the chart to tweak the nginx configmap. This can be useful to add/remove headers before requests are sent to the zammad pods.

  - fixes #69 

#### Checklist
- [x] Chart Version bumped


zammad/templates/configmap-nginx.yaml
```diff
@@ -42,6 +42,9 @@ data:
             proxy_set_header Connection "Upgrade";
             proxy_set_header CLIENT_IP $remote_addr;
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            {{ if .Values.envConfig.zammad.nginx.websocketExtraHeaders }}
+                {{ .Values.envConfig.zammad.nginx.websocketExtraHeaders | nindent 12 }}
```
monotek, line 46: Imho this should look like:

```
            proxy_set_header  {{ .Values.envConfig.zammad.nginx.websocketExtraHeaders }}
```

zammad/templates/configmap-nginx.yaml
```diff
@@ -50,6 +53,9 @@ data:
             proxy_set_header Host $http_host;
             proxy_set_header CLIENT_IP $remote_addr;
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            {{ if .Values.envConfig.zammad.nginx.extraHeaders }}
+                {{ .Values.envConfig.zammad.nginx.extraHeaders | nindent 12 }}
```
monotek, line 57: Imho this should look like:
```
             proxy_set_header  {{ .Values.envConfig.zammad.nginx.extraHeaders }}
```

zammad/values.yaml
```diff
@@ -55,6 +55,15 @@ envConfig:
...
       readinessProbe: true
+
+      # string with nginx config to set headers on the /ws location
+      # websocketExtraHeaders: |
+      #   proxy_set_header Header1 "Header Value";
+
+      # string with nginx config to set headers on the / location
+      # extraHeaders: |
+      #   proxy_set_header Header2 "Header2 Value";
+
```
monotek, line 66: Commented out stuff looks bad.
Please only comment out the value example but add an empty key.

monotek (changes requested):

zammad/templates/configmap-nginx.yaml
cabillman: I need to modify more than a single header so that will not work for my intended use case. Would you prefer I make the value a list and iterate over the items?

zammad/values.yaml
cabillman: I modeled this on: https://github.com/zammad/zammad-helm/blob/1fe1b50f7a687ba3178bc83bc5a944747059a8c4/zammad/values.yaml#L69

Are you suggesting I remove line 63, add an "extraHeaders: {}" and keep 64+65 as an example?

monotek: yes

cabillman:
I think this most recent revision gets close the the requested changes. I used a list because I need to modify more than one header.

monotek (changes requested):
Please merge master and raise chart version to 3.1.0

cabillman:
Chart version has been bumped.

monotek (approved):

PR #108, opened 2021-08-28
introduce rbac and psp support

guyguy333:

This PR adds support for RBAC and PSP

None

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 4.1.0
+version: 4.2.0
 appVersion: 4.1.0
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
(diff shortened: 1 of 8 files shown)


monotek (changes requested):
Not sure if adding psp makes still sense, as they are already deprecated: https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

guyguy333:
> Not sure if adding psp makes still sense, as they are already deprecated: https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

Indeed, they're deprecated. However, many clusters are still using it. For eg, 1.21 is not stable branch of GKE yet. I think it makes sense to enable people to use PSP and to remove it after K8S API remove (1.25 if I remember right)

monotek (changes requested):
OK, lets add it, but with the following changes:
* Add check for kubernetes version and disable if >= 1.25 is detected.
* disable serviceaccount and rbac by default too, as they are only needed for psp at the moment

guyguy333:
> OK, lets add it, but with the following changes:
> 
> * Add check for kubernetes version and disable if >= 1.25 is detected.
> * disable serviceaccount and rbac by default too, as they are only needed for psp at the moment

I updated the PR with your required changes. Thanks.

monotek (approved):

PR #112, opened 2021-10-09
refactor: storageClassName field in values.yaml and statefulset spec

### LINKED ISSUE #96: Allow empty strings pvc storageClassName
In some situations, it coud be useful to define pvc storageClassName as empty string (*not* null) to disable dynamic provisioning.

Right now, it's not fully possible in this chart, and the way of doing it may vary depends on the chart/subchart:

PostgreSQL (using the storageClass "-" de facto standard) :
```
postgresql:
  ...
  persistence:
    ...
    storageClass: "-"
```

Elasticsearch (using full "volumeClaimTemplate" block) :
```
elasticsearch:
  ...
  volumeClaimTemplate:
    ...
    storageClassName: ""
```

Zammad itself:
```
persistence:
  ...
  storageClass: ""
```

In this last example, zammad is using half of the de facto standard :)
First, storageClass is used instead of storageClassName (which, in my opinion, is a good thing).
But then, the special value "-" is not handled, and worst, if storageClass is an ampty string, it leads to no storageClassName at all because of the "with" usage right here: https://github.com/zammad/zammad-helm/blob/master/zammad/templates/statefulset.yaml#L298

zoomoid:

This PR converges the notation for the storageClassName field with those from for example bitnami, due to the outlined cases in #96 by @nervo.

  - fixes #96

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 4.2.1
+version: 5.0.0
 appVersion: 4.1.0
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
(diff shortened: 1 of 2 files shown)


monotek:
superseeded by: https://github.com/zammad/zammad-helm/pull/115

PR #123, opened 2022-02-01
Maintenance: Raise dependency versions and container versions

MrGeneration:

- Elasticsearch version has been updated to 7.16.3 (log4j relevant)
- raised chart dependencies due to security updates of upstream containers

--

I did my best to track changes in between the version our chart currently lives on to the chart versions I added.
Changes mainly were internal / default behaviors that, in my understanding, do not affect Zammad containers.

Other chart changes are security fixes in up stream containers (Debian).
This affects:
* memcached (now on latest possible chart)
* postgresql (I only updated minors; options to upgrade to latest possible should be checked)
* redis (I updated up to a version I thought I could tell nothing big changed. The next version looked like big changes that would require verifications - I stopped at that point)

If you need anything else let me know @monotek .

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 6.0.3
+version: 6.0.4
 appVersion: 5.0.3
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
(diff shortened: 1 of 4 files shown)


zammad/Chart.yaml
```diff
@@ -17,17 +17,17 @@ maintainers:
 dependencies:
   - name: elasticsearch
     repository: https://helm.elastic.co
-    version: 7.16.2
+    version: 7.16.13
```
monotek, line 20:
```suggestion
    version: 7.16.3
```

monotek (changes requested):

PR #125, opened 2022-03-17
Raise helm-chart to Zammad 5.1

MrGeneration:

This PR raises the helm-charts Zammad version to 5.1.0-4 which is the latest available docker version.

--

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 6.0.5
+version: 6.1.0
 appVersion: 5.1.0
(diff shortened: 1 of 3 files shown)


zammad/Chart.yaml
```diff
@@ -1,10 +1,10 @@
 apiVersion: v2
 name: zammad
-version: 6.0.4
-appVersion: 5.0.3
+version: 6.0.5
```
monotek, line 3: i would go up one minor version here too as this is no bugfix.
```suggestion
version: 6.1.0
```

monotek (changes requested):

monotek (approved):

PR #136, opened 2022-06-28
Raise version to Zammad 5.2

MrGeneration:
Update to Zammad 5.2

--

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: zammad
-version: 6.6.0
-appVersion: 5.1.1
+version: 6.7.0
+appVersion: 5.2.0
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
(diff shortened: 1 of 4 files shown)


mgruner:
@MrGeneration looks good to me, but I'm not familiar with Helm at all yet.

monotek (changes requested):
We would likely need to change the scheduler to background-services too, likei n https://github.com/zammad/zammad-docker-compose/pull/282

Needs to be done aroutnd: https://github.com/zammad/zammad-helm/blob/main/zammad/templates/statefulset.yaml#L266

MrGeneration:
Very good point sorry.
I'm still not having a 100% overview of helm charts and was expecting the container to be responsible for that. 🙏

monotek:
> Very good point sorry.
> I'm still not having a 100% overview of helm charts and was expecting the container to be responsible for that.

Normally you would have been right to asume that :)

But as the containers have some docker-compose specific stuff (like the ready checks) we overwrite the entrypoint here and define an own one.

monotek (approved):

PR #151, opened 2022-11-03
Add option to configure NGINX's trusted proxies

### LINKED ISSUE #150: NGINX is not trusting any reverse proxy
**Is this a request for help?**: Yes

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

**Version of Helm and Kubernetes**:
Helm 3.10.1
Kubernetes 1.25.3

**What happened**:
Zammad shows the reverse proxies' IP addresses instead of the users'.

**What you expected to happen**:
Zammad shows the users' real IP address.

**How to reproduce it** (as minimally and precisely as possible):
Standard install

**Anything else we need to know**:
This is down to the NGINX container in Zammad's StatefulSet not trusting any IP addresses and the lack of ability to configure any.

NGINX offers `set_real_ip_from` in ngx_http_realip_module. This could be used to mark the reverse proxies/cluster network as trusted.

I'm not sure what a good default would look like. Both using the reverse proxies' IP addresses and trusting too many IP addresses sounds bad to me, in a way.
Trusting every IP and using a NetworkPolicy may not work in every environment, as some network plugins do not support them.

Either way, I'll propose a PR to allow the configuration of `set_real_ip_from` to allow the configuration without disruption to existing deployments.

References:

- [Volume mount](https://github.com/zammad/zammad-helm/blob/ce0b275cdb32f5b427ab819444b4439b1f1c4183/zammad/templates/statefulset.yaml#L203-L204)
- [Volume definition](https://github.com/zammad/zammad-helm/blob/ce0b275cdb32f5b427ab819444b4439b1f1c4183/zammad/templates/statefulset.yaml#L355-L357)
- [ConfigMap definition](https://github.com/zammad/zammad-helm/blob/ce0b275cdb32f5b427ab819444b4439b1f1c4183/zammad/templates/configmap-nginx.yaml)

juliusrickert:

Adds the `zammadConfig.nginx.trustedProxies` configuration option which allows trusted proxies to be defined.

- fixes #150

Doesn't trust any proxy by default.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 6.7.1
+version: 6.8.0
 appVersion: 5.2.1
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
--- a/zammad/README.md
+++ b/zammad/README.md
@@ -52,6 +52,7 @@ The following table lists the configurable parameters of the zammad chart and th
 | `zammadConfig.memcached.enabled`            | Use Memcached dependency                         | `true`                          |
 | `zammadConfig.memcached.host`               | Memcached host                                   | `zammad-memcached`              |
 | `zammadConfig.memcached.port`               | Memcached port                                   | `11211`                         |
+| `zammadConfig.nginx.trustedProxies`         | Configure `set_real_ip_from` to trust proxies    | `[]`                            |
 | `zammadConfig.nginx.websocketExtraHeaders`  | Additional nginx headers for ws location         | `[]`                            |
 | `zammadConfig.nginx.extraHeaders`           | Additional nginx headers for / location          | `[]`                            |
 | `zammadConfig.nginx.knowledgeBaseUrl`       | Value of custom URL for knowledge base           | `""`                            |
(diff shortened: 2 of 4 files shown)


monotek (commented):
Do we need to configure this together with: https://github.com/zammad/zammad-helm/blob/main/zammad/values.yaml#L133 ?

So maybe add some note?

juliusrickert:
@monotek I've been wondering about that too.

But it is not an issue.

The Ingress [forwards to the Service](https://github.com/zammad/zammad-helm/blob/ce0b275cdb32f5b427ab819444b4439b1f1c4183/zammad/templates/ingress.yaml#L56-L57) which [selects the pod(s) of the StatefulSet](https://github.com/zammad/zammad-helm/blob/ce0b275cdb32f5b427ab819444b4439b1f1c4183/zammad/templates/service.yaml#L15) and forwards to the [`http` port](https://github.com/zammad/zammad-helm/blob/ce0b275cdb32f5b427ab819444b4439b1f1c4183/zammad/templates/service.yaml#L13) which the [NGINX container listens on](https://github.com/zammad/zammad-helm/blob/main/zammad/templates/statefulset.yaml#L195).

The NGINX container itself forwards to the Rails containers which are colocated in a Pod.

P.S. The colocation of the NGINX container with the `railsserver` and `websocket` containers makes sense to me. But why is the `scheduler` colocated in that Pod too?

juliusrickert:
I can definitely add a note about why it's not an issue, if you so desire :)

monotek:
No, a note seems not necessary.
The containers are in the same pod as they need to share the pvc.
See: https://github.com/zammad/zammad/issues/1601

monotek (approved):

PR #177, opened 2023-03-21
openshift unprivileged and with arbitrary UIDs and GIDs

klml:
#### What this PR does / why we need it

Adds values for openshift as discussed in #160

#### Which issue this PR fixes

- fixes #160

#### Special notes for your reviewer

tested on our testcluster

```
$ oc get po
NAME                                READY   STATUS    RESTARTS   AGE
postgresql-75cfcbb447-vtntk         1/1     Running   0          28h
zammad-0                            4/4     Running   0          2m46s
zammad-elasticsearch-master-0       1/1     Running   0          2m46s
zammad-memcached-7fd99d488f-qxqxh   1/1     Running   0          2m46s
zammad-redis-master-0               1/1     Running   0          2m46s
$ curl -s http://zammad-zammad.apps.capx.example.de/ | grep title
  <title>Zammad Helpdesk</title>
  <div class="splash-title">Loading…</div>
```

#### Checklist

- [x] Chart Version bumped
- [x] Upgrading instructions are documented in the README.md

PR #185, opened 2023-06-05
helm supported null

### LINKED ISSUE #160: [FR]: Arbitrary UID and GID, instead of 1000
Version of Helm and Kubernetes:

* Kubernetes Version: v1.24.6+5157800
* helm version: v3.9.4

## What happened

We want to run zammad on Openshift (<4.10). For Openshift UID (and GID) must be in an [arbitrary range](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids).
Currently zammad depends on UID 1000, so I get 

```
Warning   FailedCreate        statefulset/zammad                                       create Pod zammad-0 in StatefulSet zammad failed error:
 pods "zammad-0" is forbidden: unable to validate against any security context constraint:
 [provider "anyuid": Forbidden: not usable by user or serviceaccount, provider restricted:
 .spec.securityContext.fsGroup: Invalid value: []int64{1000}: 1000 is not an allowed group, 
spec.initContainers[0].securityContext.runAsUser: Invalid value: 1000: must be in the ranges: [1000890000, 1000899999], 
spec.initContainers[1].securityContext.runAsUser: Invalid value: 1000: must be in the ranges: [1000890000, 1000899999], ....
```

## What you expected to happen

zammad user UID and GID can be controlled by the Kubernetes /Openshift-Cluster

This feature request is even more far-reaching than [Change zammad user UID and GID from 1000 to 1111 #305](https://github.com/zammad/zammad-docker-compose/issues/305)

klml:

Improves the "magical" solution from https://github.com/zammad/zammad-helm/pull/177  (`"# must be empty"`) with the [helm supported null](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key)

- fixes #160 better

#### Checklist

- [ ] Chart Version bumped
- [x] Upgrading instructions are documented in the README.md

### DIFF
--- a/zammad/README.md
+++ b/zammad/README.md
@@ -45,24 +45,24 @@ This is relevant to **EFS** for AWS users, as well.
 To run OpenShift unprivileged and with [arbitrary UIDs and GIDs](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids):
 
 - Add the extraRsyncParams `--no-perms --omit-dir-times`.
-- Set `securityContext` and `zammadConfig.initContainers.zammad.securityContext.runAsUser` empty (not empty string `""` or empty map `{}`).
+- [Delete the default key](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key)  `securityContext` and `zammadConfig.initContainers.zammad.securityContext.runAsUser` with `null`.
 - Disable if used:
   - also `podSecurityContext` in all subcharts.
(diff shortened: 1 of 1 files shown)


klml:
@monotek fyi

I don't know which [zammad VERSION](https://github.com/zammad/zammad/blob/develop/VERSION) to bump on (`6.0`, `6.1`, `6.1.0`, ect)

monotek:
Change chart version to 8.3.1

monotek (approved):

PR #197, opened 2023-07-08
mount var volume to storage dir too

monotek:
#### What this PR does / why we need it

- mount var volume too /opt/zammad/storage too
- set postgres init readOnlyRootFilesystem to false
- update zammad to 6.0.0-66
- update dependency charts
- extend upgrading docs a bit

#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes 
  - #191 
  - #196
  - https://github.com/zammad/zammad-helm/issues/200
  - https://github.com/zammad/zammad-helm/issues/201
 
#### Special notes for your reviewer

- .

#### Checklist

[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]

- [ ] Chart Version bumped
- [ ] Upgrading instructions are documented in the README.md

PR #202, opened 2023-07-18
Ensure the volume names are consistent across configurations

### LINKED ISSUE #201: Existing volumeclaim creates volume with wrong name
https://github.com/zammad/zammad-helm/blob/3f7da1e5683041693bc0b76215d48537a584d8a3/zammad/templates/statefulset.yaml#L417

Hi,

This seems to be an oversight but when using an existing volumeClaim the volume is missing the `-var` suffix which causes the container to not start.

MTRNord:

-  It ensures that no matter what kind of volume/persistence setting you run it has the same name. Otherwise the stateful set is invalid

- fixes #201

- .

#### Checklist

- [ ] Chart Version bumped
- [ ] Upgrading instructions are documented in the README.md


monotek:
not needed. see: https://github.com/zammad/zammad-helm/pull/197

Issue #212, opened 2023-09-07
Unable to restart due `elasticsearch-init` container


johannes-darms:
After a cluster restart zammad is unable to restart, since the `elasticsearch-init` container does not complete without an error code. The logs [1]  show an error associated with a read only file system (`/opt/zammad/lib/static_assets.rb:177:in `initialize': Read-only file system @ rb_sysopen - /opt/zammad/public/assets/images/02be6b7d09f0d0eed30936e7b9ec676b.png (Errno::EROFS)`). The manifest [2] does not list a mount for `/opt/zammad/public/`. 

```
❯ kubectl logs -f zammad-demo2-0 -c elasticsearch-init
/opt/zammad/lib/static_assets.rb:177:in `initialize': Read-only file system @ rb_sysopen - /opt/zammad/public/assets/images/02be6b7d09f0d0eed30936e7b9ec676b.png (Errno::EROFS)
	from /opt/zammad/lib/static_assets.rb:177:in `open'
	from /opt/zammad/lib/static_assets.rb:177:in `sync'
	from /opt/zammad/config/initializers/logo.rb:11:in `block in <main>'
	from /usr/local/bundle/gems/activesupport-6.1.7.4/lib/active_support/callbacks.rb:427:in `instance_exec'
	from /usr/local/bundle/gems/activesupport-6.1.7.4/lib/active_support/callbacks.rb:427:in `block in make_lambda'
	from /usr/local/bundle/gems/activesupport-6.1.7.4/lib/active_support/callbacks.rb:198:in `block (2 levels) in halting'
	from /usr/local/bundle/gems/activesupport-6.1.7.4/lib/active_support/callbacks.rb:604:in `block (2 levels) in default_terminator'
	from /usr/local/bundle/gems/activesupport-6.1.7.4/lib/active_support/callbacks.rb:603:in `catch'
	from /usr/local/bundle/gems/activesupport-6.1.7.4/lib/active_support/callbacks.rb:603:in `block in default_terminator'
	from /usr/local/bundle/gems/activesupport-6.1.7.4/lib/active_support/callbacks.rb:199:in `block in halting'
	from /usr/local/bundle/gems/activesupport-6.1.7.4/lib/active_support/callbacks.rb:512:in `block in invoke_before'
	from /usr/local/bundle/gems/activesupport-6.1.7.4/lib/active_support/callbacks.rb:512:in `each'
[...]

Issue #200, opened 2023-07-13
Can not add Logo at setup or in settings


rhizoet:
**Is this a request for help?**: yes

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
BUG REPORT

**Version of Helm and Kubernetes**: 
Kubernetes: 1.26.6
Helm: 3.12.0
Zammad-helm: 9.0.4

**What happened**:
When I install Zammad and get to the part where I can select a logo, the following error message comes up:
```
Read-only file system @ rb_sysopen - /opt/zammad/public/assets/images/2e8e695017d62bf4d9cbb01703ac55c1.png
```
I have also already set `readOnlyRootFilesystem: false` for all init containers. However, this does not change anything.

**What you expected to happen**:
I assume that I can upload a logo and this will be added to Zammad. There should be no error message that this is a `read-only fs`.

**How to reproduce it** (as minimally and precisely as possible):
Install the current helm chart and try to add a logo during setup. Or in the settings when setup is finished.

**Anything else we need to know**:
![Screenshot_20230713_113854](https://github.com/zammad/zammad-helm/assets/58997736/ff0ef066-05f7-4705-b722-0aa2e668cc24)

rhizoet:
Oh, the Zammad Pod does not come up again if I had done some changes and the GUI states, that I have to restart all the services.

```
Back-off restarting failed container postgresql-init in pod zammad-0_helpdesk(2e98ac3b-66ea-4d5f-b8b2-7b1ab772c321)
```

monotek:
see: https://github.com/zammad/zammad-helm/pull/197

rhizoet:
I'm always get the same error even with the new version of the chart. Can you have a look what's up with the images?

rhizoet:
@monotek can you reopen the issue because it's not fixed?

alexkorotysh:
Any updates with broken logo?

monotek:
will be fixed in: https://github.com/zammad/zammad-helm/pull/215

PR #214, opened 2023-09-16
Added missed mount for init container

alexkorotysh:
The problem is init containers don't have possibility for checking logo and any files

- Missed mounting for init container

- issue: https://github.com/zammad/zammad-helm/issues/212
- issue: https://github.com/zammad/zammad-helm/issues/200

### DIFF
--- a/zammad/templates/statefulset.yaml
+++ b/zammad/templates/statefulset.yaml
@@ -62,6 +62,8 @@ spec:
           volumeMounts:
             - name: {{ template "zammad.fullname" . }}-var
               mountPath: /opt/zammad/var
+            - name: {{ template "zammad.fullname" . }}-var
+              mountPath: /opt/zammad/storage
         {{- end }}
         - name: zammad-init
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
@@ -99,6 +101,8 @@ spec:
           volumeMounts:
             - name: {{ template "zammad.fullname" . }}-var
               mountPath: /opt/zammad/var
+            - name: {{ template "zammad.fullname" . }}-var
+              mountPath: /opt/zammad/storage
             - name: {{ template "zammad.fullname" . }}-init
               mountPath: /docker-entrypoint.sh
               readOnly: true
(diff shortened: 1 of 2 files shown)


klml:
@mgruner we are running in the same problem with

> E, [2023-09-20T04:04:01.548417#9-6180] ERROR -- : Unable to send Ticket.find(290).search_index_update_backend backend: #<Errno::EROFS: Read-only file system @ dir_s_mkdir - /opt/zammad/storage>


This PR would solve our problem

klml:
@alexkorotysh  could you also please bump the [Chart Version](https://github.com/alexkorotysh/zammad-helm/blob/mount-for-init-container/zammad/Chart.yaml#L3) its a requirement from [zammad](https://github.com/zammad/zammad-helm/blob/main/.github/PULL_REQUEST_TEMPLATE.md#checklist)

alexkorotysh:
@klml Done with dump version of helm chart

zammad/templates/statefulset.yaml
```diff
@@ -99,6 +101,8 @@ spec:
           volumeMounts:
             - name: {{ template "zammad.fullname" . }}-var
               mountPath: /opt/zammad/var
+            - name: {{ template "zammad.fullname" . }}-var
```
mgruner, line 104: Is it intended that you use the same name `...-var` twice?

zammad/templates/statefulset.yaml
```diff
@@ -187,6 +193,8 @@ spec:
           volumeMounts:
           - name: {{ template "zammad.fullname" . }}-var
             mountPath: /opt/zammad/var
+            - name: {{ template "zammad.fullname" . }}-var
```
mgruner, line 196: Did you really want to indent this?

zammad/templates/statefulset.yaml
alexkorotysh: I just followed your code as you already have 
https://github.com/zammad/zammad-helm/blob/c9c243e36150743db002ed51a98bed9d10919c8e/zammad/templates/statefulset.yaml#L408

zammad/templates/statefulset.yaml
alexkorotysh: Fixed

mgruner:
@alexkorotysh I don't think this will solve the original issues about custom logo errors. These files get written to /opt/zammad/public/assets, which your MR does not change.

Since Zammad 6.1 is released now, updating zammad-helm to it would solve the issue without mount point changes - it no longer tries to write in `public/assets`.

monotek:
superseeded by: https://github.com/zammad/zammad-helm/pull/215

PR #232, opened 2023-11-19
fix persistence.enabled: false

pavels:
When `persistence.enabled` value is set to `false`, due to wrong indentation Helm will fail with following error: `error converting YAML to JSON: yaml: line 402: did not find expected key`


mgruner:
@pavels thanks a lot. Could you please bump the chart version to 10.0.5?

monotek (approved):

PR #238, opened 2023-12-06
Switch to Zammad 6.2

mgruner:

- Switches to Zammad 6.2
- This also switches from commit-level Zammad image reference to a floating value that will update within the minor release automatically. That's the recommended behaviour which is also active for package installations. It can still be changed locally.

#### Checklist

- [x] Chart Version bumped
- [x] Upgrading instructions are documented in the README.md

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: zammad
-version: 11.0.0
-appVersion: '6.2'
+version: 10.2.0
+appVersion: 6.2.0-1
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
 icon: https://raw.githubusercontent.com/zammad/zammad-documentation/main/images/zammad_logo_600x520.png
--- a/zammad/values.yaml
+++ b/zammad/values.yaml
@@ -1,8 +1,15 @@
 image:
   repository: ghcr.io/zammad/zammad
-  # if not set appVersion field from Chart.yaml is used as default
+  # If not set, appVersion field from Chart.yaml is used as default.
+  # appVersion points to a fixed version. You are responsible to update this to newer patch level versions yourself.
+  # Alternatively, you can also use floating versions that will give you automatic updates:
+  # tag: "6.2"     # all patchlevel updates
+  # tag: "6"       # including minor updates
+  # tag: "latest"  # all updates of stable versions, including major
+  # tag: "develop" # bleeding-edge development version
+  # If you want to use a floating version, you should also set pullPolicy: Always
   tag: ""
-  pullPolicy: Always
+  pullPolicy: IfNotPresent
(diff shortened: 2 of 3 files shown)


fliebe92 (approved):

zammad/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: zammad
-version: 10.1.1
-appVersion: 6.1.0-24
+version: '11.0.0'
+appVersion: '6.2'
```
monotek, line 4: i would prefer to use it like this, to prevent unwanted updates, even if its only patchlevel.

```suggestion
appVersion: 6.2.0
```

zammad/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: zammad
-version: 10.1.1
-appVersion: 6.1.0-24
+version: '11.0.0'
```
monotek, line 3: can we remove the quotes here?
```suggestion
version: 11.0.0
```

zammad/values.yaml
```diff
@@ -2,7 +2,7 @@ image:
   repository: ghcr.io/zammad/zammad
   # if not set appVersion field from Chart.yaml is used as default
   tag: ""
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
monotek, line 5: pullPolicy

```suggestion
  pullPolicy: IfNotPresent
```

monotek (changes requested):

zammad/Chart.yaml
mgruner: That's precisely what we intend to change here (prepared by https://github.com/zammad/zammad/pull/4920). We can choose between two evils here:

1. Don't update automatically. This means that docker and k8s installations stay on outdated patchlevel versions of Zammad and don't receive updates unless somebody submits a PR.
2. Update within patchlevels. This causes automatic updates, but only within patch level updates.

We think that the latter option is the better one, as it is a good compromise between stability and being up-to-date without maintenance burden. Having a software with bugs that are in reality already solved is not what admins should want.

Can you elaborate your reasons against this change, please?

zammad/Chart.yaml
mgruner: Yes, removed.

zammad/Chart.yaml
mgruner: This is only about changing the *default* behaviour. It's still possible to use a fixed version, or even a very loose one like `:stable` or `:latest`.

monotek: Usually you don't want to use flloating tags in production environments for sevaral reasons. Most important ones are unpredictability of image content and the risk of breaking changes. 

Even if zammad version is only a patch lavel change the whole docker container is rebuild so also changes in the os can happen.

So i see it exactly the other way around. 
If you want to use a flaotaing tag it should be an opt in.
Never the default.

mgruner: We follow your advice, @monotek

monotek (approved):

Issue #231, opened 2023-11-14
[FR] customization settings persistent file 


klml:
All customization settings (e.g. `Setting.set('storage_provider', "File")`) and [Advanced customization settings](https://docs.zammad.org/en/latest/admin/console/hidden-settings.html) shall be managed in a persistent file (e.g. `custom-settings.rb`) and applied on every deployment.

## cant use

* the [autoWizard](https://github.com/zammad/zammad-helm/blob/main/zammad/values.yaml#L301-L337), its only for inital setup
* [additional init containers](https://github.com/zammad/zammad-helm/blob/main/zammad/values.yaml#L375), cant set the needed dynamic env like `DATABASE_URL`, `REDIS_URL`  (see [zammad-railsserver](https://github.com/zammad/zammad-helm/blob/main/zammad/templates/statefulset.yaml#L274))

## Workaround

Use a persistent file (like `custom-settings.rb`) with kubectl or in your pipeline

`kubectl --namespace zammad exec -it zammad-0 --container=zammad-railsserver rails c < commands.rb`

monotek:
@mgruner what do you think about this?
adding a configmap which mounts such file seems doable.

I guess we could also add some "custom commnad" to: https://github.com/zammad/zammad-helm/blob/main/zammad/templates/configmap-init.yaml#L33 ?

klml:
@mgruner thank you very much 👍 but It stil does not work for me fup #234

PR #242, opened 2023-12-12
Openshift owned tmp

klml:

remove initContainer (for openshift uid owned /tmp for attachments upload) and use customInit from #231 in examples

## Verbosity

I am not sure if this verbosity 37b5a3171d6b82bc4e3f9387fe3e97b5ea7d8c76 is appreciated, we can remove this as you like.

#### Checklist

- [x] Chart Version bumped, its not a functional, but documantary change

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 10.2.0
+version: 10.2.1
 appVersion: 6.2.0-1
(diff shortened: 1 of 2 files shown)


klml:
>   ✖︎ zammad => (version: "10.2.0", path: "zammad") > chart version not ok. Needs a version bump! 

do you still want it to be just a change to the documentation, a version bump?

mgruner (dismissed):
Thank you @klml. Please also add a patchlevel version bump. Even this change probably changes the content of the published chart, so we should bump the version.

mgruner (approved):

mgruner:
👏

PR #243, opened 2023-12-18
Fixes: #236 - Running Zammad with replicas > 1

### LINKED ISSUE #236: Running zammad with `replicas > 1`
railsserver and nginx should be able to run with [replica >1](https://github.com/zammad/zammad-helm/blob/main/zammad/values.yaml#L360) to make the deployment highly available.

Websocket and scheduler (mainly [background-worker.rb](https://github.com/zammad/zammad/blob/develop/script/background-worker.rb)) should only be allowed to run once in the environment, to avoid double processing.

Imho there is a conflict of goals in the StatefulSet zammad.

## proposed solutions

* Separate websocket and scheduler into their own StatefulSet. (state of the art)
* conditional to limit containers zammad-scheduler and zammad-websocket only to the first replica (mitigation)

mgruner:

- Splits up the chart from one `StatefulSet` into 4 `Deployments` and one `Job`.
  - The nginx and railsserver `Deployments` are freely scalable, the scheduler and websocket must remain at `replicas: 1`
  - The `Job` will be re-created on any chart update (via uuid in the name) and run the migrations. Deployments will fail until migrations are executed.
- Add custom TMP dir handling to all deployment pods. Not just the railsserver needs the ability to create tmpfiles, but also the scheduler (e.g. image resizing for incoming emails processing). Use a consistent set of ENVs and the TMP volume for all Zammad deployment pods.
- Refactor some code into helper templates to reduce redundancy / improve maintainability.

#### Depends on

- [x] https://github.com/zammad/zammad/issues/5022
- [x] Zammad version releases which contains https://github.com/zammad/zammad/issues/5022

- fixes #236 

Open questions / issues:
- The persistentVolumeClaim will need `ReadWriteMany` access from now on. This may complicate deployments.
- Also, upgrading will involve manual intervention in case the default PVC of the `StatefulSet` was used, because a new PVC will be created.
- What about the sidecars? I don't see a place right now where to put them.

#### Checklist

- [x] Chart Version bumped

```[tasklist]
#### Open Tasks
- [x] Refactor VolumePermission handling, add custom code configuration
- [x] Check if pod affinity etc. should be specified more granular
- [ ] Wait for confirmation from LHM
```


mgruner:
@klml @monotek this is not finished yet, but I would like to ask if you want to try this and provide first feedback. Works here with nginx and railsserver scaled to `3`:

```
NAME                                                     READY   STATUS      RESTARTS   AGE
zammad-elasticsearch-master-0                            1/1     Running     0          3h33m
zammad-init-244dc656-5965-4a7b-b08f-97bf28b0fe33-hx545   0/4     Completed   0          2m14s
zammad-memcached-95c5bfd6b-lstns                         1/1     Running     0          3h33m
zammad-nginx-7f9ff54f86-56x5r                            1/1     Running     0          2m14s
zammad-nginx-7f9ff54f86-h4k6l                            1/1     Running     0          2m14s
zammad-nginx-7f9ff54f86-nn8jw                            1/1     Running     0          3h28m
zammad-postgresql-0                                      1/1     Running     0          3h33m
zammad-railsserver-58b64bb55d-267g2                      1/1     Running     0          2m14s
zammad-railsserver-58b64bb55d-9gzh9                      1/1     Running     0          8m14s
zammad-railsserver-58b64bb55d-bfndb                      1/1     Running     0          2m14s
zammad-redis-master-0                                    1/1     Running     0          3h33m
zammad-scheduler-7ddbcf4b4c-nffst                        1/1     Running     0          3h28m
zammad-websocket-86dcd556d7-5ng7s                        1/1     Running     0          3h28m
```

klml:
@mgruner Thank you very much. Openshift is working perfectly. We are still doing a few technical tests, but it looks very good.

HA works flawlessly. I just simulated a data center failure and we survived it without any measurable downtime ;)

Moving the inits to the k8s job is great, makes the rollout so much faster and also more stable, as I don't necessarily have to restart nginx and rails during a helm update.

zammad/templates/deployment-railsserver.yaml
```diff
@@ -0,0 +1,93 @@
...
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- with .Values.zammadConfig.initContainers.volumePermissions.securityContext }}
+          securityContext:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            {{- include "zammad.volumeMounts" . | nindent 12 }}
+      {{- end }}
```
mgruner, line 43: I am not entirely sure about this part, but it seems to be required to fix up the volume permissions for the container (so that `tmp/` is not world-writable).

zammad/templates/pvc.yaml
```diff
@@ -0,0 +1,21 @@
...
+    requests:
+      storage: {{ .Values.persistence.size | quote }}
+  {{- if .Values.persistence.storageClass }}
+  {{- if (eq "-" .Values.persistence.storageClass) }}
+  storageClassName: ""
+  {{- else }}
+  storageClassName: "{{ .Values.persistence.storageClass }}"
+  {{- end }}
+  {{- end }}
+{{ end }}
```
mgruner, line 21: Is this the right way to handle the volume, if no external claim is provided?
It will now use `ReadWriteMany` by default, and the underlying `PersistentVolume` has a `ReclaimPolicy` of `Delete`. This means the data will be lost when the application is uninstalled.

An alternative might be to always require an external claim to be provided when provisioning the application. This would force the admins to make their choice first.

zammad/values.yaml
```diff
@@ -399,26 +397,26 @@ initContainers: []
...
+#   #       value: "xxxxxxxxxxxx"
+#   #     - name: SYNC_DIR
+#   #       value: "/opt/zammad"
+#   #     - name: AWS_SYNC_BUCKET
+#   #       value: "some-backup-bucket"
+#   #     - name: AWS_SYNC_SCHEDULE
+#   #       value: "0 * * * *"
+#   #   volumeMounts:
+#   #     - name: help-zammad
+#   #       mountPath: /opt/zammad
```
mgruner, line 419: What should we do with the sidecars?

monotek:
sorry for the delay. 
i did not use the computer for the last 4 weeks :D
i hope i can find some time next week...

mgruner:
> sorry for the delay. i did not use the computer for the last 4 weeks :D i hope i can find some time next week...

Thanks for the update @monotek. Looking forward to your feedback!

zammad/templates/service.yaml
```diff
@@ -13,3 +13,4 @@ spec:
       name: http
```
monotek, line 13: i would call the file service-nginx to be in par with the other services

zammad/templates/statefulset.yaml
```diff
@@ -1,482 +0,0 @@
-apiVersion: apps/v1
```
monotek, line 1: As mentioned above this might still be needed if "attachments to storage option" is used.

zammad/values.yaml
```diff
@@ -357,15 +359,11 @@ persistence:
   ##
   # existingClaim:
   accessModes:
-    - ReadWriteOnce
+    - ReadWriteMany
```
monotek, line 362: see comment in Chart.yaml

zammad/values.yaml
monotek: Not entirly sure. I guess it dependes on the uscase.
Maybe for now add it to the scheduler first as it has a fixed replica of 1, so the containers are not run multiple times. Lets see than what the user feedback is.

Needs to be mentioned in the upgrading instructions too.

monotek (changes requested):

mgruner:
@monotek and I had a very productive call about this matter. We agreed that it is the right direction and will bring a major improvement for Zammad users on k8s.

However, there is an important consideration we need to make here. Zammad 6.2 currently requires `/opt/zammad/var` to be writable. This would mean we will require `ReadWriteMany` storage for all users with this change. That is not an good option on certain cloud providers / environments as @monotek already outlined.

Therefore we propose the following procedure and intermediate steps:

- Put this PR on hold for the moment.
- Add a new minor version to zammad-helm which adds support for S3 storage (based on the current `StatefulSet`). This will allow users to migrate their stored files from FS storage to S3, as a preparatory step for the laters witch to `Deployment`s.
- We will also implement a change in Zammad that modifies the way unprocessable mails are handled, so that the `var/` folder will not be used any more. We will try to implement this for Zammad 6.3, but I can't make promises here.
- Once the previous change in Zammad is implemented and released, we can rebase this PR, and eventually release it. Using `ReadWriteMany` storage would then be entirely optional, and require an `existingVolumeClaim` provided by the administrator. This also avoids the potential data loss issue on chart uninstallation.

zammad/templates/pvc.yaml
mgruner: We will change this so that an external claim is required.

zammad/templates/statefulset.yaml
mgruner: This should be solved by requiring an external claim.

klml:
Despite the fact that the branch is running, we had the problem today that a manual rake zammad:searchindex:rebuild runs into the problem like #212 ;)

klml:
> Using ReadWriteMany storage would then be entirely optional, and require an existingVolumeClaim provided by the administrator. This also avoids the potential data loss issue on chart uninstallation.

Perfekt for us. 👍

mgruner:
> Despite the fact that the branch is running, we had the problem today that a manual rake zammad:searchindex:rebuild runs into the problem like #212 ;)

I looked into this and found an issue with a missing entry point in the elasticsearch-init job container. This should be fixed by 409937d0debd7419ad652b445715f7d428de3769. Can you let me know if this helps, or otherwise send details about the error, please? It cannot be the same issue as in #212 because the StaticAssets handling is no longer present in Zammad 6.2.

zammad/values.yaml
mgruner: @monotek please see d03e08e75f2c06cce7a0f91c714359eaaaceabc9. I added a separate, optional deployment `zammad-sidecars` which has `replicas: 1` fixed and is only created if there are any sidecars present. Would you agree to this solution?

zammad/templates/deployment-sidecars.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.sidecars }}
```
monotek, line 1: the naming seems off.
a sidecar would be a container running in the same pod.

mgruner: Do you have a better suggestion? It's a POD running in the same cluster, with access to the volumes etc.
I used it to keep it compatible in the `values.yaml`.

mgruner: With this MR, there is no need any more to run in the same POD to get access to data.

monotek: Sidecars might be used for other stuff than accesing zammad data (for example adding external auth).
Therfore adding an deployment would break current functionality.
Depending on the usecase a sidecar, volume & volumemount config for every single deployment might be needed.

mgruner: Thanks for the explanation. I moved the sidecar settings now to be specific to the deployments in 47a8fdd3dd7d953ed5a7468eb0ad8098a5e15b15, can you check please?

zammad/values.yaml
mgruner: Moved sidecar settings to the deployments.

zammad/templates/service.yaml
mgruner: Renamed the service to `zammad-nginx`.

mgruner:
@klml @monotek the recent commit drops the creation of an internal PVC, and requires an `externalClaim` to be provided - but only IF `File` storage is used. Also, configuration for this moved from `persistence` to `zammadConfig.storageVolume`.

@klml this will not work any more correctly with Zammad 6.2 as it has no volume for `var/` any more.

zammad/values.yaml
mgruner: Internal PVC dropped.

zammad/README.md
```diff
@@ -104,6 +104,17 @@ Open your browser on <http://localhost:8080>
...
 ## Upgrading
 
+### From chart version 10.x to 11.0.0
+
+- The `StatefulSet` was split up into separate `Deployments`.
+  - `replicas` can be set independently now for `zammad-nginx` and `zammad-railsserver`, allowing free scaling and HA setup for these.
+  - For `zammad-scheduler` and `zammad-websocket`, `replicas` is fixed to `1` as they may only run once in the cluster.
+- The `initContainers` moved to a new `zammad-init` `Job` now which will be run on every `helm upgrade`.
+- Storage requirements changed: Zammad now requires a `ReadWriteMany` volume that is shared in the cluster.
+  - If you already had one via `persistence.existingClaim` before, it should just keep working.
```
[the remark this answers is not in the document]
mgruner: Hint added.

klml:
@mgruner thanks for leting me know. We had this branch only on dev-environment, and wait for 6.3

monotek (dismissed):
@mgruner 
Nice, that the "var" dir issue is fixed :+1: 
From the code it looks good to me now :)

Had no time to test it by myself though.

monotek (approved):

monotek (approved):

mgruner:
Sorry for the noise @monotek @klml. I found that the upgrading instructions should be slightly more precise regarding the `storageVolume`, and also changed the tests to cover this scenario as well. I guess it's useful to include this in the current PR.

monotek (approved):

klml:
@mgruner it works perfekt with an new database. But I have problems with an existing instance, and our storage_provider "File", we are using not the S3 but a ReadWriteMany PVC. 

elasticsearch-init had many `Unable to send Ticket.find(158).search_index_update_backend backend:` which leads to file problems. I try to get this next week

klml:
@mgruner [choosing-the-storage-provider](https://github.com/zammad/zammad-helm/blob/b7d4ec924d240a1524d818b3572e51ca548c2fb8/zammad/README.md#choosing-the-storage-provider)  says `externalVolumeClaim`, but I have only an `existingClaim`.

What do you mean with `externalVolumeClaim`?

mgruner:
> @mgruner [choosing-the-storage-provider](https://github.com/zammad/zammad-helm/blob/b7d4ec924d240a1524d818b3572e51ca548c2fb8/zammad/README.md#choosing-the-storage-provider) says `externalVolumeClaim`, but I have only an `existingClaim`.
> 
> What do you mean with `externalVolumeClaim`?

This probably refers to the previous state where the `StatefulSet` was present. I need to update this before merging, good point.

But first I'd like to understand if we have an issue with using the `existingClaim` as you indicated. It would be great if you could provide more details on this.

klml:
> it works perfekt with an new database. 

I must correct myself. zammad runs, but you cant upload to filesystem.

openshift needs the possibility to [use an openshift uid owned /tmp for attachments upload](https://github.com/zammad/zammad-helm/blob/main/zammad/README.md?plain=1#L106).

With [mkdir -pv /opt/zammad/tmp && chmod -v +t /opt/zammad/tmp](https://github.com/zammad/zammad-helm/pull/243/files#diff-75e781a3c90eba57f718da43eef2c89d4d3cdd51c2de74f97d028896a0a7d656R115) I get

```
failed to change mode of '/opt/zammad/tmp' from 2777 (rwxrwsrwx) to 3777 (rwxrwsrwt)
chmod: changing permissions of '/opt/zammad/tmp': Operation not permitted
```

But with `/opt/zammad/tmp/rails` this works

```
mkdir: created directory '/opt/zammad/tmp/rails'
mode of '/opt/zammad/tmp/rails' changed from 2755 (rwxr-sr-x) to 3755 (rwxr-sr-t)
```

But then uploading a image, I get 500 and 

```
system temporary path is world-writable: /tmp
/tmp is world-writable: /tmp
. is not writable: /opt/zammad
2024-04-22 10:39:38 +0000 Read: #<ArgumentError: could not find a temporary directory>

```

although the TMP is set:
```

$ oc get deployment zammad-railsserver -o yaml | grep TMP -A 1
        - name: TMP
          value: /opt/zammad/tmp/rails
```

mgruner:
> I just saw (and would likely change) that the podspec is the same for all pods, as it comes from a template or is only defined in a variable once for all pods.
> 
> Things like pod annotations might need to differ and also the pod affinitys and topologyspreadconstraints are likely not the same for every pod.

That would be feature request, as these values are defined on the top level in the `Values`. This PR does not change behaviour there and I would like to not increase it further in complexity.

> Also not sure about the container securitycontexts. Can these still be configured seperatly for every pod / container?

Yes. If you have a look at `zammad.containerSpec`, you will see that it gets the `containerConfig` passed and is able to read and use its values.

> I guess we could also use static containernames. Makes stuff like logging easier and independent from the helm release name.

Yes, I will change this!

mgruner:
> I guess we could also use static containernames. Makes stuff like logging easier and independent from the helm release name.

After checking again, I need to ask: can you explain this please? We already use static container names. Only the pod names are dynamic because they are controlled by the deployments, and this is probably by design. Did I miss something here?

mgruner:
@klml the reason for your error is probably that the `customInit` code runs only once in the job, but not in the deployments again. And the `tmpDir` is not shared between the pods. I will look for a solution here.

mgruner:
@klml can you please have a look at https://github.com/zammad/zammad-helm/pull/243/commits/f3bfc0bf7f5ebc049a02af8c6e53d07313ba0b49? This should fix the issue. It makes the volume-permissions container's command configurable, so that you can replace it with something that works in OpenShift. I would suggest to use `/opt/zammad/tmp/tmp` now to have a writable directory in the tmpDir volume. 

Please let me know if this works and if the updated description for OpenShift in the Readme is correct now.

monotek (approved):
> After checking again, I need to ask: can you explain this please? We already use static container names. Only the pod names are dynamic because they are controlled by the deployments, and this is probably by design. Did I miss something here?

Sorry, i was confused because the "chart" var is used in the container name. Example: 
https://github.com/zammad/zammad-helm/blob/b7d4ec924d240a1524d818b3572e51ca548c2fb8/zammad/templates/deployment-nginx.yaml#L28
The var is "zammad" all the time and should not change. 
Therefore lets keep it as it is.

For the rest i'm also ok with keeping it for now and change later, if somebody complains :)

mgruner:
> The "/opt/zammad/tmp/tmp" dir does look a bit weird. Can we use "/opt/zammad/var/tmp" again?

I don't think so, as there is no `var/` any more, and in a readonly FS we cannot create one. Besides that, using a subdirectory in a `tmpDir` should be much faster, as it is not a shared mount point.

Alternatively @klml you could try not modifying the `tempdir`, but instead using `zammadConfig.tmpDirVolume.emptyDir.medium: memory` as described in the comment. It would be faster and probably work permission wise without a custom subdirectory.

monotek:
I would prefer the in memory workaround too, so we can just use "/opt/zammad/tmp".
If this is no option lets go with ugly "/opt/zammad/tmp/tmp".

klml:
Great `zammadConfig.tmpDirVolume.emptyDir.medium: memory` works!  ;)

And I removed my `mkdir -pv /opt/zammad/tmp/tmp && chmod -v +t /opt/zammad/tmp/tmp` from customInit.

and I run with
```
    volumePermissions:
      enabled: false
```

I tested this on

* existing instance (startversion 10.3.4)
* fresh database

klml:
@mgruner ingress/routes were missing, because the ingress still listened to the dynamic `{{ $fullName }}`, but the svc has changed to `zammad-nginx`

```
index f4b36e3..2b5223c
--- a/zammad/templates/ingress.yaml
+++ b/zammad/templates/ingress.yaml
@@ -49,7 +49,7 @@ spec:
             backend:
               {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion }}
               service:
-                name: {{ $fullName }}
+                name: zammad-nginx
```

mgruner:
@klml absolutely awesome, thank you. Can you please check 06823636734b6e4b0bb415e01dc801021ee5bc0b? There I applied your Ingress patch and updated the OpenShift documentation in the Readme. Is this the correct and final state that we should merge now?

klml:
@mgruner works now oob ;) let me get the approval from my the functional department, then we can merge.

klml:
@mgruner looks good to me! thank you very much for this 🙏

klml:
> I guess we could also use static containernames. Makes stuff like logging easier and independent from the helm release name.

I agree and opend #265

PR #245, opened 2024-01-23
Fixes: #244 - Add support for S3 storage

### LINKED ISSUE #244: Add support for S3 storage
Zammad 6.2 was released with support for S3 as a file storage backend. This is probably the best to use in k8s environments.

We want to add support for it in the helm chart:

- [ ] Add https://artifacthub.io/packages/helm/bitnami/minio as a new dependency chart
- [ ] Allow it to be configured or turned off
- [ ] Allow usage of external S3 services
- [ ] Configure Zammad to use it by default for new installations
- [ ] Add configuration instructions
- [ ] Add instructions on how to migrate existing FS storage content to S3

We can probably release this as a new minor version. It prepares a migration path for #243.

mgruner:

- fixes #244 

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ zammad/charts
 zammad/Chart.lock
 zammad/index.yaml
 zammad/requirements.lock
+values.yaml
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -16,7 +16,7 @@ maintainers:
 dependencies:
   - name: elasticsearch
     repository: https://charts.bitnami.com/bitnami
-    version: 19.10.3
+    version: 19.17.0
     condition: zammadConfig.elasticsearch.enabled
   - name: memcached
     version: 6.5.4
@@ -30,3 +30,7 @@ dependencies:
     version: 17.11.7
     repository: https://charts.bitnami.com/bitnami
     condition: zammadConfig.redis.enabled
+  - name: minio
+    version: 13.2.0
+    repository: https://charts.bitnami.com/bitnami
+    condition: zammadConfig.minio.enabled
--- a/zammad/values.yaml
+++ b/zammad/values.yaml
@@ -81,6 +81,13 @@ zammadConfig:
     host: zammad-memcached
     port: 11211
 
+  minio:
+    # enable/disable minio chart dependency
+    enabled: true
+
+    #
+    # externalS3Url: https://user:pw@external-minio-service/bucket
+
   nginx:
     trustedProxies: []
     extraHeaders: []
@@ -451,6 +458,20 @@ memcached:
     # limits:
     #   cpu: 100m
     #   memory: 128Mi
+
+# settings for the memcached subchart
+minio:
+  auth:
+    rootUser: zammadadmin
+    rootPassword: zammadadmin
+
+    # Use existing secret for credentials details (auth.rootUser and
+    # auth.rootPassword will be ignored and picked up from this secret).
+    # The secret has to contain the keys root-user and root-password)
+    # existingSecret: minio-credentials
+
+  defaultBuckets: zammad
+  disableWebUI:   false
 
 # settings for the postgres subchart
 postgresql:
--- a/zammad/templates/_helpers.tpl
+++ b/zammad/templates/_helpers.tpl
@@ -105,3 +105,16 @@ redis secret name
 {{ template "zammad.fullname" . }}-{{ .Values.secrets.redis.secretName }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+minio access URL
+*/}}
+{{- define "zammad.S3Url" -}}
+{{- with .Values.zammadConfig.minio.externalS3Url -}}
+{{ . }}
+{{- else -}}
+{{- if .Values.zammadConfig.minio.enabled -}}
+http://zammadadmin:zammadadmin@{{ template "zammad.fullname" . }}-minio:9000/zammad?region=zammad&force_path_style=true
+{{- end -}}
+{{- end -}}
+{{- end -}}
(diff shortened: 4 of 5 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ zammad/charts
 zammad/Chart.lock
 zammad/index.yaml
 zammad/requirements.lock
+values.yaml
```
monotek, line 7: this seems wrong :)

zammad/Chart.yaml
```diff
@@ -16,17 +16,21 @@ maintainers:
...
-    version: 12.6.4
+    version: 12.12.10
     repository: https://charts.bitnami.com/bitnami
     condition: zammadConfig.postgresql.enabled
   - name: redis
-    version: 17.11.7
+    version: 17.17.1
     repository: https://charts.bitnami.com/bitnami
     condition: zammadConfig.redis.enabled
+  - name: minio
```
monotek, line 33: can we keep alphabetical sorting of dependencies please?

zammad/Chart.yaml
```diff
@@ -16,17 +16,21 @@ maintainers:
...
+    version: 13.2.0
+    repository: https://charts.bitnami.com/bitnami
+    condition: zammadConfig.minio.enabled
```
monotek, line 36: dow we need this by default as long var is not using it?
imho by default zammad should use the postgres db so this would not be needed?

zammad/README.md
```diff
@@ -13,7 +13,7 @@ This directory contains a Kubernetes chart to deploy [Zammad](https://zammad.org
 This chart will do the following:
 
 - Install Zammad statefulset
-- Install Elasticsearch, Memcached, PostgreSQL & Redis as requirements
+- Install Elasticsearch, Memcached, PostgreSQL, Redis & Minio as requirements
```
monotek, line 16:
```suggestion
- Install Elasticsearch, Memcached, PostgreSQL, Redis & Minio (optional) as requirements
```

zammad/README.md
```diff
@@ -32,14 +32,6 @@ See [Customizing the Chart Before Installing](https://helm.sh/docs/intro/using_h
 helm show values zammad/zammad
 ```
 
-### Important note for NFS filesystems
```
monotek, line 35: is this not true anymore?
in this version s3 and fs storage would be possible right?
so somebody could still use nfs and the info would be still helpfull?

zammad/README.md
```diff
@@ -78,6 +70,12 @@ memcached:
   containerSecurityContext:
     enabled: false
 
+minio:
+  podSecurityContext:
```
monotek, line 74: why do we disable podSecurityContext & podSecurityContext?

zammad/README.md
```diff
@@ -104,6 +102,39 @@ Open your browser on <http://localhost:8080>
 
 ## Upgrading
 
+### From chart version 10.x to 10.3.0
```
monotek, line 105: Normaly the upgrade infos are only used for breaking changes.
I would put these infos along the config in the values.yaml.

zammad/README.md
monotek: should this config not go to bottom of the page, to the subachrt config section?

zammad/values.yaml
```diff
@@ -81,6 +81,13 @@ zammadConfig:
     host: zammad-memcached
     port: 11211
 
+  minio:
+    # enable/disable minio chart dependency
+    enabled: true
```
monotek, line 86: dow we need it by default already?
see question above in chart.yaml.

monotek (changes requested):

.gitignore
mgruner: Why should this be wrong? It's for a top-level custom values.yaml file that the user might locally place for their customizations. It's not for the commited `zammad/values.yaml`. That seems correct to me.

zammad/Chart.yaml
mgruner: Of course, I kept it in other places, but not here. Changed.

zammad/Chart.yaml
mgruner: It is not required by default, that is true. I can disable it by default for now, but I need to think about it if that's really what we want for production systems.

mgruner: Turned off by 6205f2249c6ecfffe6947be13fd23eeb294ee7c6.

zammad/README.md
mgruner: I believe this is not true anymore, because Zammad now requires redis to be present, and if redis is available, then the web socket server will use it by default for storage. So the issue should be solved.

zammad/values.yaml
mgruner: Changed in 6205f2249c6ecfffe6947be13fd23eeb294ee7c6.

zammad/README.md
mgruner: Please have a look at the context. This change is in the OpenShift section, where I adapted it in the same way like the other entries that were already there. Should be correct IMHO.

zammad/README.md
mgruner: I will check this.

mgruner: I moved this to a new section below `Configuration` and changed it slightly.

zammad/README.md
```diff
@@ -1,48 +1,93 @@
...
+- Cluster with at least 4GB of free RAM
 
 ## Installing the Chart
 
 To install the chart use the following:
 
 ```console
 helm repo add zammad https://zammad.github.io/zammad-helm
-helm upgrade --install zammad zammad/zammad --namespace zammad
+helm upgrade --install zammad zammad/zammad
```
mgruner, line 31: I dropped the namespace for simplicity. Otherwise the command fails unless the namespace is created before.

monotek (approved):

PR #254, opened 2024-02-26
Improve CI testing coverage

mgruner:

- Improve testing by adding a `full` scenario to chart-testing which tests secrets.
  
#### Checklist

- [x] Chart Version bumped


monotek (commented):
I still don't feel to good about using the lookup function.

This could fail i some kubernetes installations, if permissions are not given.

It's also not possible to debug that with `helm template`.

`helm template` is also sometimes used to deploy manifests from a ci pipeline to a cluster.
In this case it would also fail...

mgruner:
@monotek what's your idea for the proper encoding of the passwords then? I don't know of another way.

monotek:
I don't have the time to dive into it now.
For the time being just don't use special chars.

mgruner:
@monotek @t-shehab could you review, please? Now we have another approach without using `lookup`.

mgruner:
@t-shehab @monotek I did some more changes to improve test coverage. `chart-testing` will now test a `default` configuration as well as a new `full` configuration which uses custom secrets.

zammad/templates/_helpers.tpl
```diff
@@ -63,54 +63,77 @@ Create the name of the service account to use
...
+- name: AUTOWIZARD_JSON
 {{- if .Values.secrets.autowizard.useExisting -}}
-{{ .Values.secrets.autowizard.secretName }}
+  valueFrom:
+    secretKeyRef:
+      key: {{ .Values.secrets.autowizard.secretKey }}
+      name: {{ .Values.secrets.autowizard.secretName }}
 {{- else -}}
-{{ template "zammad.fullname" . }}-{{ .Values.secrets.autowizard.secretName }}
+  value: {{ .Values.autoWizard.config | b64enc | quote }}
```
monotek, line 76: I don't see the improvment here.
Before we had a secret reference.
Now we have a plaintext secret.
From a security perpective, thats the less secure option, as exposure of the secret can happen more easy (logs, process lists or other containers in the same process namespace can read it for example).

monotek (changes requested):

zammad/templates/_helpers.tpl
mgruner: @monotek can you have another look, please? I reset the auto wizard handling to the previous state. Is it acceptable now? I would still want to add a test case for the auto wizard handling to the new 'full' scenario.

Can you please briefly explain what you mean by 'less secure' here? The auto wizard data still ends up in an ENV var of the container, no matter if there is an underlying secret or not. Are you referring to other kubernetes processes which might read the plaintext content via API?

monotek: It also applies for the other secrets.

Yes, its true that it's also an env var in the end. To prevent that we would need to use a volume mount of the secret instead using an env var.

With a plaintext secret stuff like helm template would still suffer, as the secret would be printed out directly.

mgruner: So what's your suggestion @monotek. Drop this entire MR (except for the test improvements) and close the bug as `wontfix`?

TBH I'm not sure if that's the right way. When people use existing secrets, everything is fine as before (there the security implication is not being able to use special characters, which is also not good). Only for local values from `values.yaml`, the behaviour would be different.

It seems strange that it should not be possible to provide URL encoded ENV vars via k8s in a proper way.

monotek: I don't now if there is a solution but i also don't have the time currently to check that out.
Sorry :(
I just would not like to see a solution introduced which might have security implications.

mgruner: All secrets changes reverted.

mgruner:
@monotek can you have another look, please? This no longer solves the percent encoding issue, but it does bring testing improvements which will be helpful in future.

mgruner:
@monotek and it has a small bugfix in the statefulset.

monotek (approved):

PR #256, opened 2024-02-26
bugfix(s3-storage-argocd): Fix s3 with existing secret for argocd

### LINKED ISSUE #255: S3 storage with existing auth not working in combination with argocd
**Is this a BUG REPORT or FEATURE REQUEST?**: Bug report

**Version of Helm and Kubernetes**:
Helm: v3.14.0
Kubernetes: v1.20

**What happened**:
When I use an existing secret for minio auth the deployment is not working through argocd.
This is because argocd does not support the helm lookup-function.

**What you expected to happen**:
To be able to deploy the chart using argocd

**How to reproduce it** (as minimally and precisely as possible):
Use the following config
```
zammadConfig:
  minio:
    enabled: true
minio:
  persistence:
    size: 32Gi
  auth:
    rootUser: zammadadmin
    rootPassword: zammadadmin
    existingSecret: zammad-minio
  defaultBuckets: zammad
  disableWebUI: false
```

**Anything else we need to know**:
/

stijnbrouwers:

Alternative approach to use existing secret for minio

- fixes #255

/

#### Checklist

- [X] Chart Version bumped
- [X] Upgrading instructions are documented in the zammad/README.md

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 10.3.2
+version: 10.3.3
 appVersion: 6.2.0-1
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 10.3.1
+version: 10.3.2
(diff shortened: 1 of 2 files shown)


mgruner (approved):
If we cannot use `lookup` as also suggested by @monotek, then your changes look good. Thank you!

stijnbrouwers:
@mgruner thanks!
Yeah, too bad the lookup apparently isn't always supported because I can see the need to do url encoding and don't know a (easy) way around it in the current setup...

mgruner:
Yes. I think I will change the Chart at some point so that values from the config can be percent encoded, but values from secrets simply can't.

PR #276, opened 2024-06-24
Fixes: #275 - A fixed / well known label on init job.

### LINKED ISSUE #275: A fixed / well known label on init job
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): feature

Hi,

we are using [cilium](https://cilium.io/) to secure network communication within our infrastructure. After upgrading to the new 12.2 'deployments' chart we noticed there is an init job doing a lot of things.

Currently we are not able to specify a proper policy for the init job, as we are not able to 'pick' it by it's labels properly. There are only the common `app.kubernetes.io/name` / `app.kubernetes.io/instance` labels, that have all pods, and some autogenerated labes. But no labels like a `app.kubernetes.io/component: zammad-init` as we have it for all other pods.

Is it possible to add a label at https://github.com/zammad/zammad-helm/blob/main/zammad/templates/job-init.yaml#L28?

We would also welcome #259 as this would also help to tighten the connection between the deployment and the policies.

Thanks & Bye,
     Chris

mgruner:

- fixes #275 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: zammad
-version: 12.2.0
-appVersion: 6.3.1-4
+version: 12.2.1
+appVersion: 6.3.1-56
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
 icon: https://raw.githubusercontent.com/zammad/zammad-documentation/main/images/zammad_logo_600x520.png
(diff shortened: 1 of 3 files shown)


monotek (approved):

mgruner:
@monotek sorry, I got it wrong first time and included the label in the configmap. It's now on the job and in the same way as for the deployments. 🤞

monotek:
Whoops, missed that in the review too :D

As we're already on it.... 
It might be useful to have a configurable label here, like for the other pods too...

mgruner:
> As we're already on it.... It might be useful to have a configurable label here, like for the other pods too...

Sorry, I don't get your point, can you explain please?

monotek:
Sorry, my bad. 
Seems we only have configurable annotations right now and that also works for the init job.

I guess adding configurable labels per pod would be out of scope here.

monotek (approved):

PR #279, opened 2024-07-08
Update chart dependencies.

mgruner:

- Update some chart dependencies

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -16,7 +16,7 @@ maintainers:
 dependencies:
   - name: elasticsearch
     repository: https://charts.bitnami.com/bitnami
-    version: 19.21.2
+    version: 19.17.6
     condition: zammadConfig.elasticsearch.enabled
   - name: minio
     version: 13.8.4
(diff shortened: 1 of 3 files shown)


zammad/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: zammad
-version: 12.2.2
```
monotek, line 3: if the dependencies change with major version i would also raise the charts major verion.

mgruner: @monotek thanks for the hint. I actually wanted to ask how we can perform a major update of the postgresql chart in future (as it requires manual work), but you already answered it. I went back to compatible updates for this change, don't want to create a major update now. Guess we'll need to collect some more breaking changes and upgrade all subcharts to the latest majors in a future release.

I also added the upgrade option for chart-testing, let's see if it works.

Could you please add my user `mgruner` as an admin to the `zammad` organisation on artifacthub?

monotek: invite for artiffacthub is out...

monotek (approved):

zammad/Chart.yaml
mgruner: Thank you so much @monotek!

PR #280, opened 2024-07-10
Add proper handling for common and pod specific labels and annotations.

### LINKED ISSUE #259: Ability to label deployed resources
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): feature

Many other helm charts have the ability to label deployed resources, which we use for our Velero backup system, for example list this:
```sh
    commonLabels:
      stackspin.net/backupSet: "zammad"
    podLabels:
      stackspin.net/backupSet: "zammad"
```
zammad does not seem to honor that, would be great if it did

mgruner:

- fixes #259

#### Checklist

- [x] Chart Version bumped
- [x] Clarify if this satisfies the requirements from #259


zammad/values.yaml
```diff
@@ -402,6 +402,12 @@ autoWizard:
   #     ]
   #   }
 
+commonLabels: {}
+  # my-label: "value"
+commonAnnotations: {}
+  # my-annotation: "value"
+podLabels: {}
```
monotek, line 409: imho thats not very flexible as you only can set same labels on all pods.
should be available per pod.
in the end would be same as commonLabels, or not?

zammad/values.yaml
```diff
@@ -402,6 +402,12 @@ autoWizard:
...
+  # my-label: "value"
 podAnnotations: {}
```
monotek, line 411: imho thats not very flexible as you only can set same annotations on all pods.
should be available per pod.
in the end would be same as commonAnnotations, or not?

monotek (changes requested):

zammad/values.yaml
mgruner: Should be fixed by a36b0b94cd0ec3a25119f61f54c47ae37360a6c7, can you check please?

zammad/values.yaml
mgruner: Should be fixed by a36b0b94cd0ec3a25119f61f54c47ae37360a6c7, can you check please?

@monotek Could you please add my user mgruner as an admin to the zammad organisation on artifacthub?

zammad/values.yaml
mgruner: @monotek the final state should allow for anything an admin can want to have regarding labels and annotations, global, global for pods, and pod specific.
Please let me know if this looks ok.

monotek (approved):

zammad/values.yaml
monotek: Nice :)

PR #286, opened 2024-08-13
Fix resource definitions

mueller-ma:

- The resource for postgres in values.yaml was invalid, as there's no top level `resource` key, but under `primary`: See https://github.com/bitnami/charts/blob/141e52fbb02909e3e9803319727e646a984f7bae/bitnami/postgresql/values.yaml#L481
- Add resources for elastic search.

None

None

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/zammad/values.yaml
+++ b/zammad/values.yaml
@@ -480,6 +480,14 @@ elasticsearch:
     heapSize: 512m
     masterOnly: false
     replicaCount: 1
+    resources: {}
+      # requests:
+      #   cpu: 50m
+      #   memory: 512Mi
+      # limits:
+      #   cpu: 100m
+      #   memory: 1024Mi
+
   # To use an existing Kubernetes secret containing the credentials,
   # remove the comments on the lines below and adjust them accordingly
   #
@@ -535,13 +543,14 @@ postgresql:
     #   userPasswordKey: postgresql-pass
     #   replicationPasswordKey: postgresql-replication-password
     #
-  resources: {}
-    # requests:
-    #   cpu: 250m
-    #   memory: 256Mi
-    # limits:
-    #   cpu: 500m
-    #   memory: 512Mi
+  primary:
+    resources: {}
+      # requests:
+      #   cpu: 250m
+      #   memory: 256Mi
+      # limits:
+      #   cpu: 500m
+      #   memory: 512Mi
 
 # settings for the redis subchart
 redis:
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 12.3.2
+version: 12.3.3
 appVersion: 6.3.1-121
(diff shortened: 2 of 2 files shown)


mueller-ma:
PS: The latest version of the Postgres chart adds resource requests and limits, so maybe you want to update to the latest version?

zammad/values.yaml
```diff
@@ -480,6 +480,14 @@ elasticsearch:
     heapSize: 512m
     masterOnly: false
     replicaCount: 1
+    resources: {}
```
mgruner, line 483: I don't see a `resources` top-level key in https://artifacthub.io/packages/helm/bitnami/elasticsearch, did I miss something?

mgruner (changes requested):
Thank you very much for your proposal. The elasticsearch config seems to be strange, can you check please? Also, this needs a minor level chart version bump.

For the subcharts, would it not be required to be able to specify `any` of the available settings that the subcharts provide, rather than just `resources` for elasticsearch and postgresql? I wonder if there's a generic way of allowing arbitrary modifications here, or if it is even possible by default.

mgruner:
Regarding the PG major version update: this would require a major version switch of this chart, because a manual data migration to the new pg cluster needs to be done. I would like to wait for the next bigger update, e.g. for Zammad 6.4 where we can update all subcharts to their newest major versions.

mueller-ma:
I think that the zammad chart should supply at least values for subcharts, when these values are required to run zammad. Other values might be omitted, but IMO it's handy for the user to get some value keys without looking at the subchart values.yaml file (The `values.yaml` file of `elasticsearch` is 2716 lines long!) and I see resources as basic settings.
I'm fine with deleting the resources from subcharts here.

Note that Bitnami added default resources to all charts. The default value for elasticsearch is the preset nano (https://github.com/bitnami/charts/blob/main/bitnami/elasticsearch/values.yaml#L2643), which is `"limits" (dict "cpu" "150m" "memory" "192Mi" "ephemeral-storage" "2Gi")` (https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_resources.tpl)

This isn't enough for my installation with 3 users and 1 ticket:
````
> k top pod -n zammad-test
NAME                                  CPU(cores)   MEMORY(bytes)   
zammad-elasticsearch-master-0         9m           929Mi 
````

zammad/values.yaml
mueller-ma: I added it  as `master.resources` here, not as top-level key.

Also see https://github.com/bitnami/charts/blob/main/bitnami/elasticsearch/values.yaml#L521

mgruner: Sorry, I missed that.

mgruner:
Alright, so the point is that we remove the default resource restrictions of the subchart by setting them to `{}`, right? That makes sense.

Would you be so kind to bump the Chart version in Chart.yaml to a new patchlevel, so that the CI will succeed? Then we can probably merge this.

mgruner (approved):
Very nice, thank you!

PR #318, opened 2025-02-05
Use deployment strategy: Recreate for scheduler and websocket deployments to avoid coexisting replicas

klml:

Currently the [scheduler](https://github.com/zammad/zammad-helm/blob/main/zammad/templates/deployment-scheduler.yaml) is [RollingUpdate (it is the default value)](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy). This leads to an small time period with 2 schedulers.

But the scheduler [may only run once per cluster.](https://github.com/zammad/zammad-helm/blob/main/zammad/templates/deployment-scheduler.yaml#L13).

Changing [deployment strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy) to `Recreate Deployment`

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 13.0.9
+version: 13.0.10
 appVersion: 6.4.1-32
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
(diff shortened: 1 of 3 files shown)


mgruner:
This is great @klml. We'd also need to apply this to the websocket server, can you please include it here?

mgruner (approved):

mgruner:
👏 thanks a lot @klml! 🚀

PR #323, opened 2025-02-12
 Turn off automatic reindexing by default, but reindex if no index is present yet.

### LINKED ISSUE #319: elasticsearch.reindex rollout default?
Currently, [elasticsearch-init starts a reindex](https://github.com/zammad/zammad-helm/blob/main/zammad/templates/configmap-init.yaml#L45) with every rollout, as “zammadConfig.elasticsearch.reindex” is set to [“true”](https://github.com/zammad/zammad-helm/blob/main/zammad/values.yaml#L76) by default.

With a large number of tickets and users, this elasticsearch.reindex leads to:

* a long time without a search function, because the reindexing starts with “Dropping indexes”
* an unstable init pod, as the resources (CPU/mem) have to be split between postgresql-init and elasticsearch-init

## possible solutions

* zammadConfig.elasticsearch.reindex false by default. I don't know why you have to reindex it with every rollout, even if elasticsearch is not changed at all.
  * optionally an additional cron job for reindexing 
* seperate elasticsearch-init to an extra init job, so that this pod can have its “own” ressources.

mgruner:

- fixes #319 

- This does not introduce new variables, but changes default behaviour: Reindexing is not always forced now, but it happens when no index is present. This should be what most users want. It can be turned off via `zammadConfig.elasticsearch.initalization = false`
- Therefore I hope that a minor release is enough.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 13.2.0
+version: 14.0.0
 appVersion: 6.5.0
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
--- a/zammad/templates/configmap-init.yaml
+++ b/zammad/templates/configmap-init.yaml
@@ -52,9 +52,18 @@ data:
         bundle exec rails r "Setting.set('es_user', '${ELASTICSEARCH_USER}'); Setting.set('es_password', '${ELASTICSEARCH_PASSWORD}')"
     fi
 
-    {{- if and .Values.zammadConfig.elasticsearch.reindex }}
+  {{ if .Values.zammadConfig.elasticsearch.reindex }}
     bundle exec rake zammad:searchindex:rebuild
-    {{ end }}
+  {{ else }}
+    echo "Checking if an elasticsearch index already exists…"
+    if bundle exec rails r "SearchIndexBackend.index_exists?('Ticket') || exit(1)"
+    then
+      echo "Elasticsearch index exists, no automatic reindexing is needed."
+    else
+      echo "Elasticsearch index does not exist yet, create it now…"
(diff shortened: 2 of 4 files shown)


mgruner:
@klml what do you think about this one? I hope we could do this without a major release.

klml:
@mgruner I'm fine with it as long as I can suppress the reindex during deploy ;)

zammad/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 13.0.11
+version: 13.1.0
```
monotek, line 3: Normally changing the default value of a chart is a breaking change.
If an updating user wants to keep the current behaviour he would not know of the change.
Therefore a new major version should be created & upgrade instructions should be added.

monotek (changes requested):

zammad/Chart.yaml
mgruner: Changed to 14.0.0, can you check again?

monotek (approved):

zammad/templates/configmap-init.yaml
```diff
@@ -56,6 +56,10 @@ data:
     bundle exec rake zammad:searchindex:rebuild
   {{ else }}
     echo "Checking if an elasticsearch index already exists…"
+
+    # Ensure ES connectivity, as SearchIndexBackend.index_exists? swallows internal errors.
+    bundle exec rails r "SearchIndexBackend.version"
+
     if bundle exec rails r "SearchIndexBackend.index_exists?('Ticket') || exit(1)"
```
monotek, line 63: is `|| exit(1)` really needed if `set - e` is used above?

mgruner: Yes, it is in the Ruby context. search_index_exists just returns true or false in Ruby, but we want it to return with non-zero exit code in case it is false, hence the exit(1) (also in Ruby context).

t-shehab (approved):

PR #326, opened 2025-03-04
remove nginx initialDelaySeconds

klml:

The [readiness-probe](https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/#readiness-probe) nginx delays accepting traffic for 30 seconds with [initialDelaySeconds](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes), but nginx is ready after 1-2 seconds, this unnecessarily delays my deploy or restart.

The `initialDelaySeconds: 30` on the readiness-probe can make sense for rails, because rails needs about 30 seconds to accept traffic, but the readiness-probe checks on port 3000 anyway. 

## Additional changes

Currently I would also suggest

* remove`initialDelaySeconds: 30` on the readiness-probe, it checks on port 3000 anyway. 
* remove [successThreshold: 1](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) Default is already to 1
* remove [periodSeconds: 10](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) Default is already to 10

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/values.yaml
+++ b/zammad/values.yaml
@@ -113,7 +113,6 @@ zammadConfig:
       httpGet:
         path: /
         port: 8080
-      initialDelaySeconds: 30
       successThreshold: 1
       failureThreshold: 5
       timeoutSeconds: 5
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 13.0.12
+version: 13.0.13
 appVersion: 6.4.1-45
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org


mgruner:
@klml thanks for your research. It looks like the readiness probe is somewhat abused to make for the long rails startup time. Wouldn't it be better to add a startup probe for that?

- startup probe: short interval, but high failure threshold to make for 30s startup time or more - this way it would complete earlier than 30s in case the stack is faster (e.g. in case of patchlevel update without migrations)
- readiness probe: short interval (perhaps defaults are fine?), no initialDelay needed
- liveness probe: same

We could choose a similar approach for all 3 affected pods: nginx, railsserver and websocket-server.

I tend to also agree that we can remove any values which don't change the defaults from k8s.

What do you think?

klml:
@martini 

> Wouldn't it be better to add a startup probe for that?

fullack

> startup probe: short interval, but high failure threshold to make for 30s startup time or more - this way it would complete earlier than 30s in case the stack is faster (e.g. in case of patchlevel update without migrations)

I used high failureThreshold: 10 (default is 3) with short periodSeconds: 4 (default is 10), so I get 40 secs. In this time even rails is running

>    readiness probe: short interval (perhaps defaults are fine?), no initialDelay needed
>    liveness probe: same

removed it

This stack works great for my openshift, nginx is running now in 5 secs (before 32) 😁  and rails still needs 30 secs, but better than 45 before, if I give it a full 4 GB of RAM even in 20 seconds.

mgruner:
Looks good at first sight! Will try it.

zammad/values.yaml
```diff
@@ -101,23 +101,22 @@ zammadConfig:
     # Maximum upload size
     clientMaxBodySize: 50M
     knowledgeBaseUrl: ""
+    startupProbe:
+      tcpSocket:
+        port: 8080
+      failureThreshold: 10
```
mgruner, line 107: This would mean we have a 40s startup limit. Can we increase these `startupProbe.failureThreshold` values to `20` please (also in the other places below), just to be on the safe side? Then it would be similar as before.

mgruner: Apart from that, it's ready to merge IMHO.

mgruner (approved):
👏

mgruner:
Thanks a lot @klml!

PR #328, opened 2025-03-19
Serialize container execution in init job for minimal resource usage

### LINKED ISSUE #322: Split job-init.yaml and put in order
Currently [job-init.yaml](https://github.com/zammad/zammad-helm/blob/main/zammad/templates/job-init.yaml) contains 1-4 containers

* postgresql-init for database migrations
* zammad-init (optional) for custom settings
* elasticsearch-init (optional) elasticsearch reindexing
* volumePermissions-init (optional) to change permissions in tmp

This leads to the following problems

* Especially postgresql-init and elasticsearch-init sometimes need more resources and it would be more efficient to run both on separate worker nodes.
*  If one of the jobs has problems and restarts, the others also start with.
* Easier to debug: when jobs fail, a specific pod fail.

I would suggest splitting these 2-4 containers or splitting them into individual pods. 

### 1:1 container pods
If you split each container to 1: 1 pod, you could resource, log and debug each one individually. But some jobs have dependencies:

* zammad-init can/should only start when the postgresql-init is ready and when volumePermissions-init (?) is ready
* elasticsearch-init only when the elasticsearch is green and when zammad-init is ready 

If you separate all container jobs into individual pods, you would have to include the relevant checks as init-containers.

### serial with init-container

But we could move the container `postgresql-init` as [init-containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/). Then `zammad-init` and `volumePermissions-init` will wait for the database to be ready. In addition, the pod uses its allocated resources serially and the container `postgresql-init` can start with the maximal resource of the pod.

elasticsearch-init should be created as an extra pod, but with an initjob that checks for the DB (`puts User.any?`) and whether ElasticSearch is green.

mgruner:

- fixes #322 

I decided for the `initContainer` based serial approach as it has the smallest resource footprint due to serial execution. There is now a new `postgresql-init-post` container which is only there because a job cannot have only `initContainers` and the number of these is dynamic depending on the configuration.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 13.0.13
+version: 13.1.0
 appVersion: 6.4.1-64
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
--- a/zammad/templates/configmap-init.yaml
+++ b/zammad/templates/configmap-init.yaml
@@ -22,6 +22,14 @@ data:
     fi
 
     echo "postgresql init complete :)"
+  postgresql-init-post: |-
+    #!/bin/bash
+    set -e
+
+    # Run a rails command to ensure the database is up and running.
+    bundle exec rails r "true"
+
+    echo "init sequence complete :)"
   zammad-init: |-
     #!/bin/bash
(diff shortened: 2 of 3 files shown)


mgruner:
@klml can you have a look please? Does this solve #322 to your satisfaction?

klml:
@mgruner i tested it, lgtm 🚀

PR #330, opened 2025-03-27
Fix version label

gjumic-optiva:

- app.kubernetes.io/version label should use .image.tag if it is used
- Use .Values.image.tag if used
- Defaults to .Chart.AppVersion

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 13.1.1
+version: 13.1.2
 appVersion: 6.4.1-64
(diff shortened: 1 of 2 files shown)


gjumic-optiva:
Hi,

This is just a suggestion :)

As per Kubernetes documentation [LINK](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/), this label value should be the **current** version of the application.

There were similar proposals on some popular Helm Charts repos 
https://github.com/bitnami/charts/issues/30151 -> Unfortunetly went stale
https://github.com/grafana/loki/issues/8077
https://github.com/VictoriaMetrics/helm-charts/issues/1766

Technically you could have both by adding 1 more custom label

gjumic-optiva:
@monotek  can we revisit this PR ?

PR #332, opened 2025-03-28
Add nodeSelector, tolerations and affinity configuration on the component level

gjumic-optiva:

- Adds additional configuration on the component level like nginx and railsserver deployments/pods

Let's say we have use case:

1. 2x k8s worker nodes
2. 2x nginx replicas
3. 2x railsserver replicas

With current release you are unable to set the podAntiAffinity rules to have 1x nginx and 1x railsserver on Node 1, and other 2 pods one Node 2
 

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 13.1.1
+version: 13.1.2
 appVersion: 6.4.1-64
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
--- a/zammad/values.yaml
+++ b/zammad/values.yaml
@@ -137,6 +137,30 @@ zammadConfig:
       # my-label: "value"
     podAnnotations: {}
       # my-annotation: "value"
+    nodeSelector: {}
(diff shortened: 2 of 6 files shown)


zammad/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 13.1.1
+version: 13.1.2
```
monotek, line 3: new feature should raise minor version

zammad/values.yaml
```diff
@@ -137,6 +137,30 @@ zammadConfig:
       # my-label: "value"
     podAnnotations: {}
       # my-annotation: "value"
+    nodeSelector: {}
```
monotek, line 140: while you're on it, can you please sort the YAML keys alphabetically?

monotek (changes requested):

zammad/Chart.yaml
gjumic-optiva: Done, marked as 13.2.0

zammad/values.yaml
gjumic-optiva: Only the ones I added?  I reused the same order as it was for common ones.

monotek: Would be nice for all of them :)

PR #338, opened 2025-04-30
Add the possibility to manually reindex Elasticsearch

### LINKED ISSUE #335: elasticsearch.reindex manually
As an k8s-admin I want to do start an elasticsearch.reindex in a seperate pod do reduce the load on the rails-server that is handling the web-requests.

<details>
<summary>Original request</summary>
After [switched off elasticsearch.reindex during rollout](https://github.com/zammad/zammad-helm/issues/319), you need the reindex occasionally manually (e.g. for [releases/6-5](https://zammad.com/de/releases/6-5)).

The following options come to mind:

* via rsh in a running rails pod. Disadvantage: I have the load on the web interface, which may even crash. Also, the rails server usually has fewer resources and quotas than a fast k8s-job that can get [Terminating quota-scope](https://kubernetes.io/docs/concepts/policy/resource-quotas/#quota-scopes). 
* Run the whole init-job once with `reindex: true`. Forces me to do a deploy to create the init-job, which means I have to do the postgres migrations and everything else is also touched. 
* an extra cronjob (default [without time](https://stackoverflow.com/questions/8324306/a-cron-job-that-will-never-execute)) which you can run as an job [`from=cronjob/`](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_job/)
* without [extra cronjob](https://stackoverflow.com/questions/60089350/how-do-i-manually-trigger-a-kubernates-job-not-a-cron-in-k8s) with the following hacks:
  * parallelism: 0 
  * suspend: true

All solutions have problems somehow. What would you suggest?
</details>

mgruner:

- fixes #335 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/templates/cronjob-reindex.yaml
+++ b/zammad/templates/cronjob-reindex.yaml
@@ -6,17 +6,13 @@ metadata:
     {{- include "zammad.labels" . | nindent 4 }}
     app.kubernetes.io/component: zammad-cronjob-reindex
   annotations:
-    description: |-
-      This cronjob never runs. It can be used to create maintenance jobs with
-      kubectl create job my-reindex-job --from=cronjob/{{ include "zammad.fullname" . }}-cronjob-reindex
     {{- include "zammad.annotations" . | nindent 4 }}
   {{- with .Values.zammadConfig.cronJob.reindex.annotations }}
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  # Never run this cronjob.
-  suspend: true
-  schedule: "@weekly"
+  suspend: {{ .Values.zammadConfig.cronJob.reindex.suspend | toYaml }}
+  schedule: {{ .Values.zammadConfig.cronJob.reindex.schedule | toYaml }}
   jobTemplate:
     spec:
       ttlSecondsAfterFinished: 300
--- a/zammad/templates/cronjob-reindex.yaml
+++ b/zammad/templates/cronjob-reindex.yaml
@@ -6,7 +6,9 @@ metadata:
     {{- include "zammad.labels" . | nindent 4 }}
     app.kubernetes.io/component: zammad-cronjob-reindex
   annotations:
-    description: "This cronjob never runs. It can be used to create maintenance jobs with kubectl create job --from=cronjob/reindex."
+    description: |-
+      This cronjob never runs. It can be used to create maintenance jobs with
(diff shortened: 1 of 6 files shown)


mgruner:
@klml @monotek do you want to have a first look? It worked correctly for me. 

Instructions are still missing.

mgruner:
> Do we still plan to split up the init jobs? If so the reindex jobspec could be used as a template for the reindex job and the reindex cronjob .

We don't plan this. Recent improvements were made in #328, which addressed the resource usage concerns.

mgruner:
@monotek @klml what do you think? From my side this would be good to go.

monotek (dismissed):

klml:
@mgruner it works on my cluster, thx 🙏

zammad/templates/cronjob-reindex.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "zammad.fullname" . }}-cronjob-reindex
+  labels:
+    {{- include "zammad.labels" . | nindent 4 }}
+    app.kubernetes.io/component: zammad-cronjob-reindex
+  annotations:
+    description: |-
+      This cronjob never runs. It can be used to create maintenance jobs with
```
klml, line 10: Could we change this to "never runs or default"

zammad/templates/cronjob-reindex.yaml
```diff
@@ -0,0 +1,57 @@
...
+      kubectl create job my-reindex-job --from=cronjob/{{ include "zammad.fullname" . }}-cronjob-reindex
+    {{- include "zammad.annotations" . | nindent 4 }}
+  {{- with .Values.zammadConfig.cronJob.reindex.annotations }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  # Never run this cronjob.
```
klml, line 17: Couldn't `suspend` and `schedule` be made configurable? We are currently considering whether we want to do a reindex on Sunday night after all

klml (changes requested):
Could we change to configurable and "never runs or default"?

zammad/templates/cronjob-reindex.yaml
mgruner: Removed

klml (approved):

monotek (approved):

PR #344, opened 2025-05-26
add support for http 1.1 in the nginx config

ash8m:

- This PR sets `proxy_http_version 1.1` in the `location /` block of the NGINX ConfigMap to ensure consistency with other location blocks (e.g., `/cable` and `/ws`) that already specify HTTP/1.1.

- This change is necessary to support integration with services that require HTTP/1.1, such as when deploying Zammad in environments like Istio-based service meshes. Without this, the default fallback to HTTP/1.0 can cause compatibility issues.

- This was discussed in the [Zammad development community forum](https://community.zammad.org/t/http-1-1-support-for-the-nginx-location-block/17759) 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 14.1.2
+version: 14.1.3
 appVersion: 6.5.0-75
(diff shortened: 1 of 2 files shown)


mgruner (approved):
Super awesome, thanks for spotting this. Waiting for CI results before merging. I'll also create a same change for Zammad's nginx config (which is also used in the docker compose stack).

PR #362, opened 2025-09-12
Introduce priorityClassName to pods

ceejaykelly:
This will allow administrators to introduce priority classes to the pods deployed by this chart, giving greater control over scheduling and pre-emption.

#### Checklist

- [X] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 15.2.5
+version: 15.3.0
 appVersion: 6.5.2-22
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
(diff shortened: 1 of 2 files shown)


zammad/templates/_helpers.tpl
```diff
@@ -273,6 +273,9 @@ imagePullSecrets:
 {{- if .Values.serviceAccount.create }}
 serviceAccountName: {{ include "zammad.serviceAccountName" . }}
 {{- end }}
+{{- if .Values.priorityClassName }}
```
monotek, line 276: We have multiple components which might have different priorities so i would see this change made separately for every pod.

zammad/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 15.2.1
+version: 15.2.2
```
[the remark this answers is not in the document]
ceejaykelly: Minor version has now been raised.

zammad/templates/_helpers.tpl
ceejaykelly: Modified the helper template, placed it all pod templates.

monotek (commented):
.

monotek (changes requested):
new option has to be added to the values.yaml too.

PR #368, opened 2025-10-08
update chart dependencies

monotek:

- .

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, …)` format, will close that issue when PR gets merged)*

- update chart dependencies
  - use bitnami oci chart repo
  - replaces bitnamilegacy with bitnamicsecure where possible
    - as only latest tag is available for bitnamisecure images the sha digest is used to fix it to a specific version
- todo: 
  - use official elasticsearch chart: https://github.com/elastic/cloud-on-k8s/tree/main/deploy/eck-stack/charts/eck-elasticsearch
  - use official minio chart: https://github.com/minio/minio/tree/master/helm/minio

- 

#### Checklist

- [ ] Chart Version bumped


greedybro:
@monotek I looked a bit at the versions of postgreSQL and it seems that starting with the 10.x release, there is no patch versions anymore. You can see all versions published [here](https://www.postgresql.org/docs/release/)

So the returning version of `psql -V` is accurate. You can also use `SELECT version();` in a psql cli to check
image

greedybro:
i'm also curious about the format of `latest@sha256:0eeb6604ad4258924850a38dfcd56697351086ede82e9c7668451c50569548b2`

why use `latest@` ? docker pull returns an `invalid reference format` when using it

monotek:
It depends how much work you want to invest.
As far as i know the cloudpirates charts are no forks of the bitnami ones so there is likely no easy update path.

PR #373, opened 2025-10-28
Switch to cloudpirates redis / memcached chart and image

### LINKED ISSUE #367: redis cve-2025-49844 Critical CVSS 10.
redis currently has a Critical security vulnerability (cvss 10).

https://redis.io/blog/security-advisory-cve-2025-49844/

I know that until [the migration away from bitnami is not finished](https://github.com/zammad/zammad-helm/issues/353) this will be difficult to address.

mgruner:

- fixes #367 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/values.yaml
+++ b/zammad/values.yaml
@@ -827,7 +827,7 @@ redis:
     # existingSecret: redis-pass
     # existingSecretPasswordKey: redis-password
   resources: {}
-  # limits: q q
+  # limits:
   #   cpu: 250m
   #   memory: 256Mi
   # requests:


mgruner:
@monotek I'm not entirely sure if this would be a good path, but the cloudpirates/redis helm chart seems to be modeled after the bitnami chart. Data migration should not be needed. We could also consider switching memcached as well. Would this require a major update?

monotek:
> [cloudnative-pg](https://artifacthub.io/packages/helm/cloudnative-pg/cloudnative-pg)

I've tried cnpg a while ago and the cluster was broken from time to time.
So currently i still use the [zalando postgres operator](https://github.com/zalando/postgres-operator).

For zammad chart usecase (single pod / no cluster) it might be ok though.
I also would give cnpg a new chance as there is now also active active support via [pgEdge](https://cloudnative-pg.io/blog/matthew-mols-pgedge/).

If we use operators now we could also switch to the [eck operator for elasticsearch](https://github.com/elastic/cloud-on-k8s/tree/main/deploy/eck-operator).

PR #389, opened 2025-12-16
feat: make init job ttlSecondsAfterFinished configurable

### LINKED ISSUE #359: Allow setting ttlSecondsAfterFinished on the init job
When using mechanism like rancher fleet - ttlSecondsAfterFinished set to 300 forces an update as the job is being deleted after 5 minutes.  The Init job is then recreated every 5-6 minutes.  It would be great if we can set that value (in my case null) in the chart as it is currently hardcoded to 300.  A combination of this with randomName would fix the resync loop issue.

simonhir:

- Allow configuration of init Job `ttlSecondsAfterFinished`

- closes #359

#### Checklist

- [x] Chart Version bumped
- ~~[ ] Upgrading instructions are documented in the zammad/README.md~~ (not required)

### DIFF
--- a/zammad/values.yaml
+++ b/zammad/values.yaml
@@ -500,7 +500,7 @@ zammadConfig:
   initJob:
     randomName: true
     # specify how long after finish pods are deleted
-    ttlSecondsAfterFinished: 3600
+    ttlSecondsAfterFinished: 300
     annotations: {}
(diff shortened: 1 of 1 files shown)


mgruner:
LGTM, what do you think @monotek?

PR #394, opened 2026-02-10
[FIX][NAMESPACE]:Add namespace injection to all Helm templates using .Release.Namespace

devlopergalxboy:

Summary of this PR
Implements namespace injection in all Helm templates using .Release.Namespace to ensure resources are deployed in the intended namespace.
Related Issues
(If applicable, reference issues in the format: closes #123)
Additional Information for Reviewers
Please let me know if further changes or clarifications are needed.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 15.6.2
+version: 15.6.3
 appVersion: 6.5.2-85
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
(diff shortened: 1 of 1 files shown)


devlopergalxboy:
> @devlopergalxboy thanks for the PR! Could you please let us know what exact problem it solves? AFAICT it's already possible to install Zammad into namespaces via `helm --namespace`.

@mgruner when we tried to deploy using rendered manifest pattern we identified that most of our resources were not being injected with the correct namespace which left us out of sync on our ArgoCD as such we thought it would be a great idea that who ever wants to use zammad with the rendered manifest pattern they can do so without having issues of namespace injection.

Yes adding the flag --namespace to the command works well.

PR #425, opened 2026-05-11
feat: add support for TLS protocol to Redis connection (rediss://)

dreglad:

- Introduces a `tls` configuration option under `zammadConfig.redis` (defaulting to `false` for backwards compatibility).
- Updates the `REDIS_URL` in `_helpers.tpl` to dynamically use the `rediss://` protocol schema when TLS is enabled, instead of hardcoding `redis://`.
- This fixes connectivity issues and enables native SSL/TLS connections when connecting Zammad to external Redis instances, or configuring Redis subchart to use TLS/SSL.

- Helm template generation was tested locally and correctly provisions `rediss://` only when `tls: true`.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 16.0.14
+version: 16.0.15
 appVersion: 7.0.1-0031
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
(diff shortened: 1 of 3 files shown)


monotek (dismissed):

zammad/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 16.0.14
+version: 16.0.15
```
mgruner: This should probably be a 16.1.0 to indicate a new feature.

dreglad: Sure

mgruner (approved):

mgruner:
Thanks and 👏  @dreglad

PR #427, opened 2026-05-15
Update fsGroupChangePolicy default on values.yaml

### LINKED ISSUE #378: Parameter `fsGroupChangePolicy: Always` causes very slow pod start
Having `fsGroupChangePolicy` set to `Always` causes all pods using the volume defined in `zammadConfig.storageVolume` to take a very long time to start, when that volume begins to grow and uses a shared FS like CephFS or NFS it becomes unfeasible. In our case Helm deploy aborts because the timeout is exceeded.

I believe this is not viable for large environments and it would be better to set it to `OnRootMismatch` by default.

References:

- https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods
- https://kubernetes.io/blog/2020/12/14/kubernetes-release-1.20-fsgroupchangepolicy-fsgrouppolicy/
- https://github.com/kubernetes/kubernetes/issues/69699

It was set by @alexkorotysh on #210

juan-ferrer-toribio:

- Changes `securityContext.fsGroupChangePolicy` default value to `OnRootMismatch`, to avoid very long start time when (big) volumes with big amount of files are used.

- fixes #378

### DIFF
--- a/zammad/values.yaml
+++ b/zammad/values.yaml
@@ -68,7 +68,7 @@ secrets:
 securityContext:
   fsGroup: 1000
   # https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods
-  fsGroupChangePolicy: Always
+  fsGroupChangePolicy: OnRootMismatch
   runAsUser: 1000
   runAsNonRoot: true
   runAsGroup: 1000


monotek:
We already have it configurable and changing the default would need a major version bump.
Imho not worth it.

mgruner:
@monotek it might be good to change this together with another major version change, e.g. switching away from more bitnami charts. It could improve general usability here.

monotek:
OK for me. 
Maybe together with PDB enabled by default.

mgruner:
Included now in #445, which already is a major update.

PR #437, opened 2026-06-17
Update to Zammad 7.0.2

mgruner:


fliebe92 (approved):

mgruner:
@monotek we just noticed that the install-chart CI tests have become extremely slow (almost 2h). Can you give us a tip how to solve this?

mgruner:
@monotek it's super strange because there is a `--timeout 900` which does seem to be ignored.

PR #458, opened 2026-08-14
Fixes #457 - Random init job name is too long

### LINKED ISSUE #457: Random init job name is too long
```plain
helm -n zammad install zammad-migration-test-with-long-name zammad 

Error: INSTALLATION FAILED: server-side apply failed for object 
zammad/zammad-migration-test-with-long-name-init-03bd3838-ef17-4ea5-b448-6754b53852c4 
batch/v1, Kind=Job: Job.batch 
"zammad-migration-test-with-long-name-init-03bd3838-ef17-4ea5-b448-6754b53852c4" 
is invalid: spec.template.labels: Invalid value: 
"zammad-migration-test-with-long-name-init-03bd3838-ef17-4ea5-b448-6754b53852c4": 
must be no more than 63 bytes
````

This effectively prevents longer names from being usable, unless `.Values.zammadConfig.initJob.randomName` is disabled.

### PR DESCRIPTION
**Pull request by mgruner**

- fixes #457 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/zammad/templates/job-init.yaml
+++ b/zammad/templates/job-init.yaml
@@ -4,7 +4,7 @@ kind: Job
 metadata:
   {{- if .Values.zammadConfig.initJob.randomName }}
   # Use a different job name on each run to ensure a new job always runs once.
-  name: {{ include "zammad.fullname" . }}-init-{{ uuidv4 }}
+  name: {{ include "zammad.fullname" . }}-init-{{ .Release.Revision }}
   {{- else }}
   name: {{ include "zammad.fullname" . }}-init
   {{- end}}
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 17.0.0
+version: 17.0.1
 appVersion: 7.1.2-0004
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
--- a/zammad/README.md
+++ b/zammad/README.md
@@ -177,6 +177,9 @@ zammadConfig:
       argocd.argoproj.io/hook: Sync
 ```
 
+Note that `randomName` is a bit of a misnomer: the job name suffix is the Helm release revision, not a random value.
+Tools that render the chart statically instead of performing a real Helm install/upgrade (Argo CD, `helm template`, `helm diff`) always see revision `1`, so with `randomName: true` the job name never changes across syncs and the job will not re-run on its own - hence the `false` + Sync Hook workaround above. Flux's helm-controller uses the Helm SDK directly and is unaffected by this.
+
 ## Maintenance Tasks
 
 If you have to run any maintenance commands inside of the Zammad stack, you can do that by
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by mgruner** [comment]
@fliebe92 could you have a look at this, please? There's still as small gap compared to the 63 character limit to the release name, but I'm accepting that. We probably also have this in other places with suffixes.

File: zammad/templates/job-init.yaml [hunk 1]
```diff
@@ -4,7 +4,7 @@ kind: Job
 metadata:
   {{- if .Values.zammadConfig.initJob.randomName }}
   # Use a different job name on each run to ensure a new job always runs once.
-  name: {{ include "zammad.fullname" . }}-init-{{ uuidv4 }}
+  name: {{ include "zammad.fullname" . }}-init-{{ .Release.Revision }}
```
fliebe92, line 7: This solves the reported case, but the remaining gap is not static — it shrinks as the revision grows, so the failure is delayed rather than removed.

`zammad.fullname` is `trunc 63`, then `-init-` (6 chars) plus the revision digits are appended, so the constraint is `len(fullname) <= 57 - digits(revision)`:

* rev 1–9: fullname <= 56
* rev 10–99: fullname <= 55
* rev 100+: fullname <= 54

That means a release with a 56-char fullname installs fine, upgrades fine nine times, and then fails on the 10th upgrade when the revision crosses to two digits. The `uuidv4` version at least failed immediately and deterministically; failing only after nine successful upgrades is a worse shape of the same bug.

Since the suffix length is known at render time, it can be closed exactly:

```gotemplate
  {{- $suffix := printf "-init-%d" (.Release.Revision | int) }}
  name: {{ printf "%s%s" (include "zammad.fullname" . | trunc (int (sub 63 (len $suffix))) | trimSuffix "-") $suffix }}
```

(The `randomName: false` branch below has the same latent overflow — 63-char fullname + `-init` = 68 — but that one is pre-existing, so feel free to leave it out of this patch.)

zammad/templates/job-init.yaml
```diff
@@ -4,7 +4,7 @@ kind: Job
 metadata:
   {{- if .Values.zammadConfig.initJob.randomName }}
   # Use a different job name on each run to ensure a new job always runs once.
```
fliebe92, line 6: Worth a doc note somewhere (`values.yaml` comment and/or the ArgoCD section of the README): `.Release.Revision` is always `1` under `helm template`, so for statically rendered manifests (ArgoCD, `helm diff`, rendered-manifest GitOps) the name never changes and the job never re-runs. Flux's helm-controller uses the Helm SDK, so it is unaffected.

Not a regression in practice — the README already tells ArgoCD users to set `randomName: false` plus the Sync hook annotation — but the failure mode flips from "loops forever creating jobs" to "silently never re-runs", and `randomName` is now a misnomer since the name is fully deterministic. Renaming the value would be breaking, so just documenting that the suffix is the release revision seems enough for 17.0.1.

Everything else looks good: revision increments on install/upgrade/rollback including failed and no-op upgrades, so the "a new job always runs once" guarantee holds, and `helm diff`/`--dry-run` output is stable now instead of churning on every render.

zammad/templates/job-init.yaml
mgruner: As mentioned in my comment above, this is acceptable at the moment. We probably have the similar issue in other places. Can be addressed later if needed.

zammad/templates/job-init.yaml
mgruner: Good catch. Docs improved.

fliebe92 (approved):

PR #460, opened 2026-08-18
Maintenance: document helm 4 workaround and activate the upgrade: true flag for chart-testing

mgruner:

- Reactivate the `upgrade: true` flag for chart-testing to cover upgrade paths.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/.github/ct.yaml
+++ b/.github/ct.yaml
@@ -5,7 +5,7 @@ chart-repos:
   - elastic=https://helm.elastic.co
 check-version-increment: true
 debug: true
-# upgrade: true # Causes issues because of low performance on default GitHub actions servers.
+upgrade: true
 target-branch: main
 namespace: zammad
 release-label: zammad
--- a/.github/ct.yaml
+++ b/.github/ct.yaml
@@ -6,6 +6,8 @@ chart-repos:
 check-version-increment: true
 debug: true
 upgrade: true
+# Needed because of CI splits value to its own job, so that default can be missing.
+skip-missing-values: true
 target-branch: main
 namespace: zammad
 release-label: zammad
--- a/.github/helm-force-conflicts-wrapper.sh
+++ b/.github/helm-force-conflicts-wrapper.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+#
+# Shadows the real `helm` binary on PATH during `ct install` to
+#   strip `--force-conflicts` for subcommands that don't understand it.
+#
+# Helm 4 defaults to Server-Side Apply, which conflicts with fields (e.g.
+# .spec.nodeSets) the ECK operator takes ownership of after install, so
+# `helm upgrade` needs --force-conflicts. But chart-testing's
+# --helm-extra-args is passed to every helm subcommand it runs (install,
+# upgrade, test, uninstall), and --force-conflicts is only accepted by
+# install/upgrade/rollback - ct has no way to scope it per subcommand
+# (helm/chart-testing#540, closed as not_planned). This wrapper strips
+# --force-conflicts for any subcommand that doesn't understand it, so it can
+# stay in --helm-extra-args without breaking `helm test`/`helm uninstall`.
+# Remove once chart-testing supports per-subcommand extra args.
+#
+# Expects $REAL_HELM to point at the actual helm binary.
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+case "${1:-}" in
+  install | upgrade | rollback)
+    exec "${REAL_HELM}" "$@"
+    ;;
+  *)
+    args=()
+    for arg in "$@"; do
+      [[ "${arg}" == "--force-conflicts" ]] || args+=("${arg}")
+    done
+    exec "${REAL_HELM}" "${args[@]}"
+    ;;
+esac
--- a/.github/helm-force-conflicts-wrapper.sh
+++ b/.github/helm-force-conflicts-wrapper.sh
@@ -21,7 +21,36 @@ set -o pipefail
 
 : "${REAL_HELM:?REAL_HELM must point at the real helm binary}"
 
-case "${1:-}" in
+# The subcommand isn't necessarily $1: Helm's global flags (--namespace,
+# --kube-context, ...) are valid before it too, e.g. `helm --namespace zammad
+# upgrade ...`. Scan for the first non-flag token, skipping the value of any
+# preceding flag that takes one. Helm's only boolean global flags are --debug
+# and --kube-insecure-skip-tls-verify (see `helm help environment`); every
+# other "-"-prefixed global flag takes a value.
+find_subcommand() {
+  local skip_next=false
+  for arg in "$@"; do
+    if [[ "${skip_next}" == true ]]; then
+      skip_next=false
+      continue
+    fi
+    case "${arg}" in
+    --debug | --kube-insecure-skip-tls-verify | -h | --help)
+      continue
+      ;;
+    -*)
+      [[ "${arg}" == *=* ]] || skip_next=true
+      continue
+      ;;
+    *)
+      echo "${arg}"
+      return
+      ;;
+    esac
+  done
+}
+
+case "$(find_subcommand "$@")" in
 install | upgrade | rollback)
   exec "${REAL_HELM}" "$@"
   ;;
--- a/.github/helm-force-conflicts-wrapper.sh
+++ b/.github/helm-force-conflicts-wrapper.sh
@@ -21,36 +21,7 @@ set -o pipefail
 
 : "${REAL_HELM:?REAL_HELM must point at the real helm binary}"
 
-# The subcommand isn't necessarily $1: Helm's global flags (--namespace,
-# --kube-context, ...) are valid before it too, e.g. `helm --namespace zammad
-# upgrade ...`. Scan for the first non-flag token, skipping the value of any
-# preceding flag that takes one. Helm's only boolean global flags are --debug
-# and --kube-insecure-skip-tls-verify (see `helm help environment`); every
-# other "-"-prefixed global flag takes a value.
-find_subcommand() {
-  local skip_next=false
-  for arg in "$@"; do
-    if [[ "${skip_next}" == true ]]; then
-      skip_next=false
-      continue
-    fi
-    case "${arg}" in
-    --debug | --kube-insecure-skip-tls-verify | -h | --help)
-      continue
-      ;;
-    -*)
-      [[ "${arg}" == *=* ]] || skip_next=true
-      continue
-      ;;
-    *)
-      echo "${arg}"
-      return
-      ;;
-    esac
-  done
-}
-
-case "$(find_subcommand "$@")" in
+case "${1:-}" in
 install | upgrade | rollback)
   exec "${REAL_HELM}" "$@"
   ;;
--- a/.github/helm-force-conflicts-wrapper.sh
+++ b/.github/helm-force-conflicts-wrapper.sh
@@ -21,14 +21,14 @@ set -o pipefail
 set -o nounset
 
 case "${1:-}" in
-  install | upgrade | rollback)
-    exec "${REAL_HELM}" "$@"
-    ;;
-  *)
-    args=()
-    for arg in "$@"; do
-      [[ "${arg}" == "--force-conflicts" ]] || args+=("${arg}")
-    done
-    exec "${REAL_HELM}" "${args[@]}"
-    ;;
+install | upgrade | rollback)
+  exec "${REAL_HELM}" "$@"
+  ;;
+*)
+  args=()
+  for arg in "$@"; do
+    [[ "${arg}" == "--force-conflicts" ]] || args+=("${arg}")
+  done
+  exec "${REAL_HELM}" "${args[@]}"
+  ;;
 esac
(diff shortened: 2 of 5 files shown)


zammad/Chart.yaml
```diff
@@ -1,6 +1,7 @@
 apiVersion: v2
 name: zammad
-version: 18.0.0
+version: 18.0.1
+# Just a test comment
```
fliebe92: Leftover debug artifact — this needs to go before merge (the PR is out of draft now). It also visually attaches to `appVersion`, which will confuse the next reader.

Removing it is safe for the CI coverage of this PR: `zammad/README.md` is still a chart-dir change, so `check-version-increment` stays satisfied by the 18.0.1 bump and `install-chart` still runs.

.github/workflows/ci.yaml
```diff
@@ -137,5 +148,6 @@ jobs:
       - name: Install additional objects for 'full' test scenario
         run: kubectl create --namespace zammad --filename zammad/ci/full-objects.yaml
 
+      # Use --force-testing because of https://helm.sh/community/hips/hip-0023/#conflicts-and-forcing.
```
fliebe92: Wrong flag name: it's `--force-conflicts`, not `--force-testing`.

.github/helm-force-conflicts-wrapper.sh
```diff
@@ -0,0 +1,34 @@
...
+#
+# Expects $REAL_HELM to point at the actual helm binary.
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+case "${1:-}" in
+install | upgrade | rollback)
+  exec "${REAL_HELM}" "$@"
```
fliebe92, line 26: `nounset` catches an *unset* `REAL_HELM`, but not an *empty* one. If `azure/setup-helm` ever renames or drops its `helm-path` output, `REAL_HELM=` silently becomes empty and every helm call in the job dies with a cryptic path error that reads like a chart failure rather than a wrapper misconfiguration.

Cheap insurance — fail with a clear message instead:

```bash
: "${REAL_HELM:?REAL_HELM must point at the real helm binary}"
```

right below the `set -o` block.

.github/helm-force-conflicts-wrapper.sh
```diff
@@ -0,0 +1,34 @@
...
+  ;;
+*)
+  args=()
+  for arg in "$@"; do
+    [[ "${arg}" == "--force-conflicts" ]] || args+=("${arg}")
+  done
+  exec "${REAL_HELM}" "${args[@]}"
```
fliebe92, line 33: Minor: expanding an empty array as `"${args[@]}"` under `nounset` only works on bash >= 4.4, so a bare `helm` with no subcommand would abort with `args[@]: unbound variable`. Irrelevant on `ubuntu-24.04` (bash 5.2), but it would bite anyone running this script on macOS' system bash 3.2. `"${args[@]+"${args[@]}"}"` or dropping `nounset` avoids it — take it or leave it.

.github/helm-force-conflicts-wrapper.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+#
+# Shadows the real `helm` binary on PATH during `ct install` to
```
fliebe92, line 3: Wording: `GITHUB_PATH` makes this shadow `helm` for the *whole job*, not just during `ct install` — the *Install ECK operator* step goes through the wrapper too (harmless: `install` passes straight through, `repo add`/`repo update` just lose a flag they never had). Worth stating accurately, since this comment is the only documentation of the hack.

.github/ct.yaml
```diff
@@ -5,7 +5,7 @@ chart-repos:
   - elastic=https://helm.elastic.co
 check-version-increment: true
 debug: true
-# upgrade: true # Causes issues because of low performance on default GitHub actions servers.
+upgrade: true
```
fliebe92, line 8: Worth documenting here (or in the PR description): this gives previous-revision coverage for **minor/patch bumps only**. ct hard-skips the old→new upgrade on a major bump, and your own earlier run in this PR shows it — at chart version 19.0.0 the log read:

```
Skipping upgrade test of "zammad 19.0.0" because: * 19.0.0 does not have same major version as 18.0.0
Testing upgrades of chart "zammad 19.0.0" relative to previous revision "zammad 19.0.0"
```

i.e. it silently degraded to upgrading the new chart onto itself. So the coverage disappears at exactly the riskiest releases — the 17.x → 18.0.0 ECK migration would not have been covered. A one-line comment stops a future maintainer over-trusting the green check.

.github/workflows/ci.yaml
```diff
@@ -137,5 +148,6 @@ jobs:
...
       - name: Run chart-testing (install)
-        run: ct install --config .github/ct.yaml --helm-extra-args '--timeout 900s'
+        run: ct install --config .github/ct.yaml --helm-extra-args '--timeout 900s --force-conflicts'
```
fliebe92, line 167: Two observations on this step, neither blocking.

1. `--force-conflicts` now applies to every install/upgrade in CI, so we'll also force through any *future* SSA field-ownership conflict, not just the known ECK one. ct gives no way to scope it, so this is the right trade-off — just a conscious one.

2. The job has no `timeout-minutes`. It now runs four install+test+upgrade+test rounds and took 15.5–18 min per k8s version in the last green run (previously ~12 min). Since runner performance was the original reason `upgrade: true` was disabled, a `timeout-minutes: 45` on the `install-chart` job would fail a hung run fast instead of letting it drift toward the 6 h default.

For the record, I verified the mechanism works end-to-end in run [32111701169](https://github.com/zammad/zammad-helm/actions/runs/32111701169) (52ff404, chart 18.1.0): green on all three kind versions, and the log shows a real `Testing upgrades … relative to previous revision "zammad 18.0.0" (ct-previous-revision…/zammad)` → `helm install` 18.0.0 → `helm test` → `helm upgrade` to 18.1.0 → `helm test` → `helm uninstall`, for both `default-values.yaml` and `full-values.yaml`. `helm test`/`helm uninstall` ran with `--force-conflicts` stripped, and `REAL_HELM` resolved to `/opt/hostedtoolcache/helm/4.2.4/x64/linux-amd64/helm`. The approach is sound.

.github/helm-force-conflicts-wrapper.sh
mgruner: Not important, accept as is.

.github/ct.yaml
mgruner: Not important, accept as is.

zammad/Chart.yaml
mgruner: Removed.

.github/workflows/ci.yaml
mgruner: Fixed, thanks.

.github/helm-force-conflicts-wrapper.sh
mgruner: Added, thanks.

.github/helm-force-conflicts-wrapper.sh
mgruner: Dropped nounset, because we explicitly check REAL_HELM now.

.github/workflows/ci.yaml
mgruner: Added timeout-minutes: 30.

fliebe92 (dismissed):

.github/helm-force-conflicts-wrapper.sh
```diff
@@ -0,0 +1,35 @@
...
+  exec "${REAL_HELM}" "$@"
+  ;;
+*)
+  args=()
+  for arg in "$@"; do
+    [[ "${arg}" == "--force-conflicts" ]] || args+=("${arg}")
+  done
+  exec "${REAL_HELM}" "${args[@]}"
+  ;;
+esac
```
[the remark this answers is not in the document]
mgruner: Changed, check again.

.github/helm-force-conflicts-wrapper.sh
```diff
@@ -0,0 +1,64 @@
...
+set -o errexit
+set -o pipefail
+
+: "${REAL_HELM:?REAL_HELM must point at the real helm binary}"
+
+# The subcommand isn't necessarily $1: Helm's global flags (--namespace,
+# --kube-context, ...) are valid before it too, e.g. `helm --namespace zammad
+# upgrade ...`. Scan for the first non-flag token, skipping the value of any
+# preceding flag that takes one. Helm's only boolean global flags are --debug
+# and --kube-insecure-skip-tls-verify (see `helm help environment`); every
```
fliebe92: Wrong pointer: `helm help environment` documents environment variables — global flags come from `helm --help`. Worth fixing, because this whitelist is the load-bearing assumption of the whole function: a boolean global flag that's *not* listed consumes the next token, so `helm --newbool upgrade …` would silently lose `--force-conflicts` and the SSA conflict would come back. It fails loudly in CI rather than passing green, so the risk is fine — but the comment should point at the authoritative source.

I exercised the function against a stub helm and it's correct for everything ct and this workflow actually emit: flag preserved for `install`/`upgrade`/`rollback` (also behind `--namespace zammad`, `--debug`, and as `upgrade --install`), stripped for `test`/`uninstall`/`repo add` (also behind `-n zammad`, `--namespace=zammad`, `--kubeconfig /path`), pass-through for `--help` and no args, and a clean abort when `REAL_HELM` is unset.

One observation and one nit: ct always puts the subcommand first (`>>> helm install …`, `>>> helm test …` in every log), as do this workflow's own `helm repo add` / `helm install elastic-operator` calls — so none of the flag-before-subcommand cases occur today, and this is ~30 lines of defensiveness in a shim whose header says to remove it once ct supports per-subcommand args. Your call whether that's worth it. And `skip_next` is `local` while `arg` isn't, so `arg` leaks to global scope (cosmetic, the script `exec`s immediately).

.github/workflows/ci.yaml
```diff
@@ -58,7 +58,6 @@ jobs:
     runs-on: ubuntu-24.04
     needs:
       - lint-chart
```
fliebe92, line 60: With `super-linter` no longer gating anything, a lint failure no longer stops the install-chart matrix — that's now ~54 min of runner time spent on a PR that can't merge anyway. Reasonable trade for faster feedback, just more pronounced since the job count doubled. Keeping `super-linter` in `install-chart.needs` only (leaving `kubeconform-chart` ungated) would get you both, if you want it.

.github/workflows/ci.yaml
```diff
@@ -94,24 +94,48 @@ jobs:
           - v1.33.7
           - v1.34.3
           - v1.35.1
+        ci-values:
```
fliebe92, line 97: This works out really well: 8–10 min per leg in the last run, down from 15–19, so `timeout-minutes: 30` now has ~3× headroom and my earlier concern about that value is moot. Total runner time is roughly unchanged too (6 × ~9 ≈ 54 min vs 3 × ~17 ≈ 51 min), so you nearly halved wall-clock for just the doubled kind/ECK setup overhead.

One thing to check before merge: the job names changed from `install-chart (v1.33.7)` to `install-chart (v1.33.7, default)`. That usually strands required status checks. Classic branch protection returns 404 on `main` and the rulesets API only shows one disabled ruleset, so from what I can see nothing pins those names — but I don't have admin visibility, so worth confirming.

.github/workflows/ci.yaml
```diff
@@ -94,24 +94,48 @@ jobs:
...
       - name: Set up chart-testing
         uses: helm/chart-testing-action@v2.8.0
 
+      - name: Isolate zammad/ci/${{ matrix.ci-values }}-values.yaml
+        # ct has no flag to pick a single ci/*-values.yaml file; it always tests every
+        # one it finds. Removing the other scenario's file here (workspace-only, never
+        # committed) makes ct test only this matrix leg's scenario. --skip-missing-values
+        # below then stops ct from erroring about that file being "removed" relative to
+        # the previous revision, which still has both.
+        run: find zammad/ci -name '*-values.yaml' ! -name '${{ matrix.ci-values }}-values.yaml' -delete
```
fliebe92, line 137: Verified this does what the comment says. In the `default` leg's log ct reports `Upgrade testing for values file ".../ci/full-values.yaml" skipped because a corresponding values file was not found in zammad/ci`, while the real 18.0.0 → 18.0.1 upgrade still ran for the default scenario, followed by the same-version round and the plain install — so `--skip-missing-values` skips only the deleted scenario, not the upgrade test itself. The `find` pattern also correctly leaves `full-objects.yaml` and `README.md` alone.

Nit: `--skip-missing-values` sits on the CLI while every other ct option lives in `.github/ct.yaml`; `skip-missing-values: true` there would be more consistent.

.github/workflows/ci.yaml
mgruner: No, thank you. Our runner minutes are free.

.github/workflows/ci.yaml
mgruner: Checked. Not relevant.

.github/workflows/ci.yaml
mgruner: Good point, moved to the .yaml file.

.github/helm-force-conflicts-wrapper.sh
mgruner: I agree and restored the previous, simple approach. Copilot was too eager here.

fliebe92 (approved):

PR #441, opened 2026-06-17
Redis sentinel sentinels k8s-service

klml:
#### What this PR does / why we need it

Removes the switch to use the vanilla k8s-service as "sentinels" when sentinel is activated.

* if redis was enabled, what is mandatory for redis-sentinel, the else (with the configurable "sentinels") could never be used.
* if you want to use sentinel, there is no reason to use the vanilla not-sentinel "sentinels"

#### Checklist

- [x] Chart Version bumped
- [ ] Upgrading instructions are documented in the zammad/README.md

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Redis Sentinel connectivity by consistently using the configured Sentinel endpoint.
  - Updated the default Redis Sentinel address to support reliable service discovery.

- **Chores**
  - Bumped the Helm chart version from 16.4.2 to 16.4.3.

PR #464, opened 2026-09-01
Redis Sentinel: more robust hostname handling + docs

mgruner:
## What this PR does / why we need it

Started as a follow-up to #441, which was opened by an external contributor to fix a legitimate-looking bug but never got updated to address CI failures. **Update after review:** @fliebe92 showed that the original premise (the plain `<release>-redis` service can't serve as a Sentinel endpoint) doesn't hold — see the review discussion below. This PR is therefore no longer a bugfix; it's a robustness/clarity improvement plus documentation for a previously undocumented feature.

### Background

#441's actual regression (confirmed) was dropping the `.Release.Name`-based branch for `REDIS_SENTINELS` and replacing it with a static list, which broke installs under any release name other than `zammad` (e.g. `ct install` in CI, which randomizes release names).

#441's stated motivation — that the bundled Redis subchart's plain `-redis` service shouldn't be used as the Sentinel target — turned out to be incorrect: when `sentinel.enabled && architecture == replication`, the subchart's plain `-redis` service *also* exposes port `26379` (`targetPort: sentinel`) via the identical pod selector as the dedicated `-redis-sentinel` service, and is explicitly annotated for Sentinel-based discovery. Zammad's Redis client also defaults a portless entry to `26379`. So `{{ .Release.Name }}-redis` (old, release-name-aware) and `{{ .Release.Name }}-redis-sentinel:26379` (this PR) currently resolve to the same endpoint.

### What this PR actually does

1. **Keeps the release-name-aware `-redis-sentinel:26379` value** — harmless, and more explicit/future-proof than relying on the plain service's incidental Sentinel-port exposure.
2. **Adds a fail-fast guard**: `zammadConfig.redis.sentinel.enabled=true` with the bundled subchart requires `redis.architecture=replication` — that's the only mode in which a Sentinel service exists at all. Previously this misconfiguration would silently produce a Sentinel hostname that resolves to nothing (or, before this PR, to a service without the Sentinel port), surfacing only as a hung rollout.
3. **Documents Redis Sentinel in the README** (bundled and external setups) — this was completely undocumented before, which is arguably the more valuable fix for the confusion behind #441.
4. Chart version bump corrected to a patch release (`18.0.4`), matching existing versioning conventions.

## Checklist

- [x] Chart Version bumped (18.0.3 → 18.0.4)
- [x] Upgrading instructions — not needed, no breaking change; new fail-fast guard only rejects an already-broken configuration

## Test plan

- [x] `helm template` with a randomized release name (mirroring `ct install`) confirms `REDIS_SENTINELS` matches the actually-created `<release>-redis-sentinel` service
- [x] `helm template` with `redis.enabled=false` confirms the external-Redis static `sentinels` list is unaffected
- [x] `helm template` with `sentinel.enabled=true` and default `architecture=standalone` now fails fast with a clear error instead of hanging
- [x] `helm lint` passes
- [x] CI (`ct install` with `full-values.yaml`, Sentinel + replication) green

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation requiring replication architecture when bundled Redis Sentinel is enabled.
  - Preserved the existing Sentinel endpoint and support for externally managed Redis and Sentinel configurations.

- **Documentation**
  - Clarified bundled and external Sentinel setup, required settings, and automatic service discovery.
  - Updated examples for disabling bundled Redis and specifying external Sentinel instances.

- **Chores**
  - Updated the Helm chart version to 18.0.4.

### DIFF
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 18.1.0
+version: 18.0.4
 appVersion: 7.1.3-0006
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
--- a/zammad/Chart.yaml
+++ b/zammad/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 18.0.3
+version: 18.1.0
 appVersion: 7.1.3-0006
 description: Zammad is a web based open source helpdesk/customer support system with many features to manage customer communication via several channels like telephone, facebook, twitter, chat and e-mails.
 home: https://zammad.org
--- a/zammad/README.md
+++ b/zammad/README.md
@@ -142,7 +142,7 @@ zammadConfig:
       enabled: true
 ```
 
-`redis.sentinel.enabled` and `redis.architecture=replication` are both required whenever `zammadConfig.redis.sentinel.enabled` is set — the chart fails fast at render time otherwise. The Sentinel hostname is derived automatically from the release name; you don't need to set `zammadConfig.redis.sentinel.sentinels`.
+When bundled Redis is enabled, `redis.sentinel.enabled` and `redis.architecture=replication` are both required whenever `zammadConfig.redis.sentinel.enabled` is set — the chart fails fast at render time otherwise. The Sentinel hostname is derived automatically from the release name; you don't need to set `zammadConfig.redis.sentinel.sentinels`.
 
 To use an **external** Redis/Sentinel setup instead, disable the bundled subchart and point at your own Sentinels:
 
--- a/zammad/README.md
+++ b/zammad/README.md
@@ -134,7 +134,7 @@ To enable Sentinel with the **bundled** Redis subchart:
 redis:
   sentinel:
     enabled: true
-  architecture: replication  # required: the Sentinel service only exists in this mode
+  architecture: replication # required: the Sentinel service only exists in this mode
 
 zammadConfig:
   redis:
@@ -147,16 +147,14 @@ zammadConfig:
 To use an **external** Redis/Sentinel setup instead, disable the bundled subchart and point at your own Sentinels:
 
 ```yaml
-redis:
-  enabled: false
-
 zammadConfig:
   redis:
+    enabled: false
     sentinel:
       enabled: true
       sentinels:
-      - my-sentinel-1:26379
-      - my-sentinel-2:26379
+        - my-sentinel-1:26379
+        - my-sentinel-2:26379
       masterName: mymaster
 ```
 
--- a/zammad/README.md
+++ b/zammad/README.md
@@ -142,7 +142,7 @@ zammadConfig:
       enabled: true
 ```
 
-`redis.architecture` must be `replication` whenever `redis.sentinel.enabled` is set — the chart fails fast at render time otherwise. The Sentinel hostname is derived automatically from the release name; you don't need to set `zammadConfig.redis.sentinel.sentinels`.
+`redis.sentinel.enabled` and `redis.architecture=replication` are both required whenever `zammadConfig.redis.sentinel.enabled` is set — the chart fails fast at render time otherwise. The Sentinel hostname is derived automatically from the release name; you don't need to set `zammadConfig.redis.sentinel.sentinels`.
 
 To use an **external** Redis/Sentinel setup instead, disable the bundled subchart and point at your own Sentinels:
 
--- a/zammad/README.md
+++ b/zammad/README.md
@@ -124,6 +124,42 @@ I, [2024-01-24T11:06:14.566513#168-5980]  INFO -- : Moved file dbaa01dd0df3a33bc
 I, [2024-01-24T11:06:14.627896#168-5980]  INFO -- : storage remove '/opt/zammad/storage/fs/e81f/fb09/c5a26/f2081/f93401a/cbe8fff/9983e56c86fccb48d17a2eb1e5900b5b'
 ```
 
+### Redis Sentinel
+
+The chart can talk to Redis through Sentinel instead of a direct connection, either using the bundled Redis subchart or an external Redis/Sentinel setup.
+
+To enable Sentinel with the **bundled** Redis subchart:
+
+```yaml
+redis:
+  sentinel:
+    enabled: true
+  architecture: replication  # required: the Sentinel service only exists in this mode
+
+zammadConfig:
+  redis:
+    sentinel:
+      enabled: true
+```
+
+`redis.architecture` must be `replication` whenever `redis.sentinel.enabled` is set — the chart fails fast at render time otherwise. The Sentinel hostname is derived automatically from the release name; you don't need to set `zammadConfig.redis.sentinel.sentinels`.
+
+To use an **external** Redis/Sentinel setup instead, disable the bundled subchart and point at your own Sentinels:
+
+```yaml
+redis:
+  enabled: false
+
+zammadConfig:
+  redis:
+    sentinel:
+      enabled: true
+      sentinels:
+      - my-sentinel-1:26379
+      - my-sentinel-2:26379
+      masterName: mymaster
+```
+
 ### Deploying on OpenShift
 
 To deploy on OpenShift unprivileged and with [arbitrary UIDs and GIDs](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids):
--- a/zammad/templates/_helpers.tpl
+++ b/zammad/templates/_helpers.tpl
@@ -266,6 +266,9 @@ Redis Variables
 {{- if .Values.zammadConfig.redis.sentinel.enabled }}
 - name: REDIS_SENTINELS
 {{- if .Values.zammadConfig.redis.enabled }}
+{{- if ne .Values.redis.architecture "replication" }}
+{{ fail "zammadConfig.redis.sentinel.enabled requires redis.architecture=replication when the bundled Redis subchart is used (redis.enabled=true); the Sentinel service is only created in replication mode." }}
+{{- end }}
   value: "{{ .Release.Name }}-redis-sentinel:26379"
 {{- else }}
   value: "{{ join "," .Values.zammadConfig.redis.sentinel.sentinels }}"
--- a/zammad/templates/_helpers.tpl
+++ b/zammad/templates/_helpers.tpl
@@ -266,8 +266,8 @@ Redis Variables
 {{- if .Values.zammadConfig.redis.sentinel.enabled }}
 - name: REDIS_SENTINELS
 {{- if .Values.zammadConfig.redis.enabled }}
-{{- if ne .Values.redis.architecture "replication" }}
-{{ fail "zammadConfig.redis.sentinel.enabled requires redis.architecture=replication when the bundled Redis subchart is used (redis.enabled=true); the Sentinel service is only created in replication mode." }}
+{{- if not (and .Values.redis.sentinel.enabled (eq .Values.redis.architecture "replication")) }}
+{{ fail "zammadConfig.redis.sentinel.enabled requires redis.sentinel.enabled=true and redis.architecture=replication when the bundled Redis subchart is used (zammadConfig.redis.enabled=true); the Sentinel service is only created in that combination." }}
 {{- end }}
   value: "{{ .Release.Name }}-redis-sentinel:26379"
 {{- else }}
--- a/zammad/templates/_helpers.tpl
+++ b/zammad/templates/_helpers.tpl
@@ -266,7 +266,7 @@ Redis Variables
 {{- if .Values.zammadConfig.redis.sentinel.enabled }}
 - name: REDIS_SENTINELS
 {{- if .Values.zammadConfig.redis.enabled }}
-  value: "{{ .Release.Name }}-redis"
+  value: "{{ .Release.Name }}-redis-sentinel:26379"
 {{- else }}
   value: "{{ join "," .Values.zammadConfig.redis.sentinel.sentinels }}"
 {{- end }}
--- a/zammad/values.yaml
+++ b/zammad/values.yaml
@@ -304,9 +304,11 @@ zammadConfig:
     tls: false
 
     sentinel:
-      enabled: false  # set to true to enable Redis Sentinel; requires redis.architecture=replication when redis.enabled=true
-      # only used when redis.enabled=false (external Redis/Sentinel); the bundled subchart's
-      # sentinel service is discovered automatically and does not use this list
+      enabled: false  # set to true to enable Redis Sentinel; with the bundled subchart
+                      # (zammadConfig.redis.enabled=true) this also requires
+                      # redis.sentinel.enabled=true and redis.architecture=replication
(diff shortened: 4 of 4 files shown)


zammad/templates/_helpers.tpl
```diff
@@ -266,7 +266,7 @@ Redis Variables
 {{- if .Values.zammadConfig.redis.sentinel.enabled }}
 - name: REDIS_SENTINELS
 {{- if .Values.zammadConfig.redis.enabled }}
-  value: "{{ .Release.Name }}-redis"
+  value: "{{ .Release.Name }}-redis-sentinel:26379"
```
fliebe92, line 272: This change is functionally a no-op — the old value already worked, so I don't think this fixes a bug.

I rendered both branches against the real subchart (`cloudpirates-io/redis` 0.34.26) with `ci/full-values.yaml` and a `ct`-style release name, and traced the consuming code:

1. `main` renders `REDIS_SENTINELS: "zammad-abc123-redis"` (no port).
2. Zammad defaults a portless entry to **26379** — `lib/zammad/service/redis.rb:29-32`: `{ host:, port: port&.to_i || 26_379 }`.
3. The plain `-redis` service *does* expose 26379 (`targetPort: sentinel`) when sentinel + replication are on, and is annotated for exactly this use: *"This service points to all Redis instances for Sentinel-based discovery / Applications should use Sentinel clients to discover the current master."*
4. Both services have byte-identical selectors (`app.kubernetes.io/instance` + `app.kubernetes.io/name`, **no** `component` filter), so `-redis:26379` and `-redis-sentinel:26379` resolve to the same set of pods.

So the real regression in #441 was solely dropping the `.Release.Name` branch — that part of the analysis was right. But its parenthetical that commit `397f837` "correctly moved the target from the plain `-redis` service to the new `-redis-sentinel` service" was never verified, and it doesn't hold. That premise carried into this PR's description. The green CI here doesn't demonstrate a fix either, since `main` passes the `full` scenario too.

I'd still merge the line — `-redis-sentinel:26379` is more explicit and doesn't lean on an implicit port default, and it stays correct if a future subchart bump adds a `component: sentinel` selector to the sentinel service. But I'd reframe the PR as a clarity change rather than a fix, and either keep #441 open or close it as "not a bug". As it stands, the release notes would tell admins with a broken Sentinel setup that their problem is solved, when they'd still be broken and would stop looking for the actual cause.

zammad/templates/_helpers.tpl
```diff
@@ -266,7 +266,7 @@ Redis Variables
 {{- if .Values.zammadConfig.redis.sentinel.enabled }}
 - name: REDIS_SENTINELS
 {{- if .Values.zammadConfig.redis.enabled }}
```
fliebe92, line 268: Separate point, but this branch is the natural place for it: the new hostname is *more* fragile than the old one in one reachable misconfiguration.

`-redis-sentinel` is created only under `and .Values.sentinel.enabled (eq .Values.architecture "replication")`, and the chart default is `redis.architecture: standalone`. Rendering with `redis.sentinel.enabled=true` but the default architecture:

```
services:  ...-redis-headless, ...-redis      ← no -redis-sentinel
REDIS_SENTINELS: "zammad-abc123-redis-sentinel:26379"   ← NXDOMAIN
```

To be fair this isn't a regression — the old value was broken there too (the plain service only exposes 6379 in standalone, so it'd fail as a protocol error rather than a DNS failure). But since this line is being touched anyway, a guard would turn a 900s hang into an immediate, readable error. There's precedent in `zammad.elasticsearchName` above:

```gotemplate
{{- if and .Values.zammadConfig.redis.enabled (ne .Values.redis.architecture "replication") }}
{{- fail "Redis Sentinel requires redis.architecture=replication and redis.sentinel.enabled=true." }}
{{- end }}
```

Related, and probably the more valuable fix for #441's underlying frustration: `grep -i sentinel zammad/README.md` returns nothing. Sentinel is completely undocumented, so there is nowhere a user learns that `zammadConfig.redis.sentinel.enabled`, `redis.sentinel.enabled` and `redis.architecture` all have to be aligned.

Minor, and correct to leave untouched since it's the external-Redis path: `values.yaml` still defaults `sentinel.sentinels` to `[zammad-redis:26379]`, which now reads as inconsistent with the bundled path's `-redis-sentinel` naming. A one-line comment would save the next reader wondering which is right.

zammad/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: zammad
-version: 18.0.3
+version: 18.1.0
```
fliebe92: This should be a patch bump — `18.0.4`.

Convention from the `Chart.yaml` history: patch for fixes (#458 "Fixes #457" → `17.0.0` → `17.0.1`, #460 → `18.0.0` → `18.0.1`), minor for features, major for subchart migrations (#445, #450). Whether this lands as a bugfix or as the clarity change I argued for in `_helpers.tpl`, neither warrants `18.1.0`.

zammad/templates/_helpers.tpl
mgruner: You're right, and I verified it against the vendored subchart: the plain `-redis` service also exposes 26379 with an identical selector under sentinel+replication, and Zammad defaults a portless entry to 26379 anyway — so the old value already worked. Reframed the PR description and title accordingly; this line is now just an explicitness/robustness improvement, not a bugfix. Keeping it per your suggestion.

zammad/templates/_helpers.tpl
mgruner: Both taken. Added a fail-fast guard for sentinel.enabled without architecture=replication (same style as zammad.elasticsearchName's fail), and added a Redis Sentinel section to the README covering both the bundled and external setups. Also added a one-line comment on the external sentinels default in values.yaml clarifying it's unrelated to the bundled path's naming.

zammad/Chart.yaml
mgruner: Fixed to 18.0.4.

zammad/README.md
```diff
@@ -124,6 +124,42 @@ I, [2024-01-24T11:06:14.566513#168-5980]  INFO -- : Moved file dbaa01dd0df3a33bc
...
+      enabled: true
+```
+
+`redis.architecture` must be `replication` whenever `redis.sentinel.enabled` is set — the chart fails fast at render time otherwise. The Sentinel hostname is derived automatically from the release name; you don't need to set `zammadConfig.redis.sentinel.sentinels`.
+
+To use an **external** Redis/Sentinel setup instead, disable the bundled subchart and point at your own Sentinels:
+
+```yaml
+redis:
+  enabled: false
```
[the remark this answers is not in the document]
mgruner: Good catch, fixed — the example now sets zammadConfig.redis.enabled=false instead of the unrelated top-level redis.enabled.

zammad/templates/_helpers.tpl
```diff
@@ -266,7 +266,10 @@ Redis Variables
...
-  value: "{{ .Release.Name }}-redis"
+{{- if ne .Values.redis.architecture "replication" }}
```
fliebe92: The guard only checks half of the subchart's condition, so the misconfiguration it was added for can still get through.

The sentinel service is created under `and .Values.sentinel.enabled (eq .Values.architecture "replication")` — but this only checks the architecture. `redis.sentinel.enabled` defaults to **false**, so it slips past:

```yaml
redis:
  architecture: replication     # redis.sentinel.enabled left at its default (false)
zammadConfig:
  redis: {sentinel: {enabled: true}}
```

```
render exit=0                                    ← guard did not fire
REDIS_SENTINELS: "rel-xyz-redis-sentinel:26379"
services:  rel-xyz-redis-headless, rel-xyz-redis-master, rel-xyz-redis   ← no -redis-sentinel
```

NXDOMAIN, i.e. exactly the hung rollout the guard is meant to turn into a readable error. Mirroring the subchart's own condition closes it. I patched this locally and confirmed it fails fast on the case above while leaving the valid bundled config, the external config, and both `ci/default-values.yaml` and `ci/full-values.yaml` rendering clean:

```gotemplate
{{- if not (and .Values.redis.sentinel.enabled (eq .Values.redis.architecture "replication")) }}
{{- fail "zammadConfig.redis.sentinel.enabled requires redis.sentinel.enabled=true and redis.architecture=replication when the bundled Redis subchart is used (zammadConfig.redis.enabled=true); the Sentinel service is only created in that combination." }}
{{- end }}
```

For fairness: the guard I suggested in the previous round had this same gap — I only checked the architecture too.

While you're in here, the message on the next line says `(redis.enabled=true)`, but the flag is `zammadConfig.redis.enabled` (`Chart.yaml`'s dependency condition). Same wrong key CodeRabbit flagged in the README example — worth fixing in both places, since this string is what a user who trips the guard actually reads, and it currently points them at a knob that doesn't exist.

zammad/README.md
```diff
@@ -124,6 +124,40 @@ I, [2024-01-24T11:06:14.566513#168-5980]  INFO -- : Moved file dbaa01dd0df3a33bc
...
+    enabled: true
+  architecture: replication # required: the Sentinel service only exists in this mode
+
+zammadConfig:
+  redis:
+    sentinel:
+      enabled: true
+```
+
+`redis.architecture` must be `replication` whenever `redis.sentinel.enabled` is set — the chart fails fast at render time otherwise. The Sentinel hostname is derived automatically from the release name; you don't need to set `zammadConfig.redis.sentinel.sentinels`.
```
fliebe92: This names the wrong flag as the trigger. The guard hangs off `zammadConfig.redis.sentinel.enabled`, not `redis.sentinel.enabled`, so as written the sentence doesn't hold:

```yaml
redis:
  sentinel:
    enabled: true
  # architecture left at its default 'standalone'
# zammadConfig.redis.sentinel.enabled left at its default false
```
→ renders fine, exit 0. No fail-fast.

Fixing the guard gap I flagged in `_helpers.tpl` makes the *combination* actually validated, but the sentence should still name the flag that triggers it — something like "`redis.sentinel.enabled` and `redis.architecture=replication` are both required whenever `zammadConfig.redis.sentinel.enabled` is set; the chart fails fast at render time otherwise."

Three similarly-named `enabled` flags interact here (`zammadConfig.redis.enabled`, `zammadConfig.redis.sentinel.enabled`, `redis.sentinel.enabled`), and that confusion is the source of both this and CodeRabbit's README finding below — so it's worth being pedantic about which one is meant in each sentence.

zammad/README.md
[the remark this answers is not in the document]
fliebe92: Confirming this one is real — I reproduced it rather than just reading it. Copying the example verbatim doesn't merely leave the bundled Redis enabled, it aborts the render outright:

```
$ helm template rel-xyz . -f readme-external-example.yaml
Error: zammadConfig.redis.sentinel.enabled requires redis.architecture=replication
       when the bundled Redis subchart is used ...
```

With the suggested fix applied it behaves as documented: `REDIS_SENTINELS: "my-sentinel-1:26379,my-sentinel-2:26379"` and zero bundled-redis templates rendered.

So the proposed diff is correct as-is. Related, in `_helpers.tpl` — the `fail` message repeats the same wrong key (`redis.enabled=true`); I've commented there separately.

zammad/templates/_helpers.tpl
mgruner: Confirmed, reproduced the exact case you described (architecture=replication, sentinel.enabled left at default false) — guard didn't fire, no -redis-sentinel service, dead REDIS_SENTINELS. Applied your patch (mirroring the subchart's own and condition) and fixed the wrong key name in the fail message too. Re-verified all 5 scenarios (your gap case now fails fast, ci/default-values.yaml, ci/full-values.yaml, the README bundled and external examples all render clean).

zammad/README.md
mgruner: Reworded per your suggestion — now names redis.sentinel.enabled and redis.architecture=replication as the actual requirement, triggered by zammadConfig.redis.sentinel.enabled.

zammad/README.md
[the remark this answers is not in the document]
mgruner: Thanks for reproducing it — that was against 9b441fe, before the zammadConfig.redis.enabled fix landed. Re-checked with the current example (helm template with the exact README yaml): renders clean, REDIS_SENTINELS resolves to the external list, zero bundled-redis templates. Also fixed the wrong key in the _helpers.tpl fail message you flagged, in the same commit as the guard-gap fix.

zammad/README.md
```diff
@@ -124,6 +124,40 @@ I, [2024-01-24T11:06:14.566513#168-5980]  INFO -- : Moved file dbaa01dd0df3a33bc
...
+    enabled: true
+  architecture: replication # required: the Sentinel service only exists in this mode
+
+zammadConfig:
+  redis:
+    sentinel:
+      enabled: true
+```
+
+`redis.sentinel.enabled` and `redis.architecture=replication` are both required whenever `zammadConfig.redis.sentinel.enabled` is set — the chart fails fast at render time otherwise. The Sentinel hostname is derived automatically from the release name; you don't need to set `zammadConfig.redis.sentinel.sentinels`.
```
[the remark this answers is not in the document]
mgruner: Applied verbatim.

zammad/values.yaml
```diff
@@ -304,7 +304,9 @@ zammadConfig:
     tls: false
 
     sentinel:
-      enabled: false  # set to true to enable Redis Sentinel
+      enabled: false  # set to true to enable Redis Sentinel; requires redis.architecture=replication when redis.enabled=true
```
fliebe92: Nit, and the only thing I have left on this PR — these two comment lines are now the last place carrying the old wording.

1. **Stale:** this line says the requirement is `redis.architecture=replication`, but the guard now also requires `redis.sentinel.enabled=true`. The README sentence was updated to cover both; this wasn't.
2. **`redis.enabled` isn't a real key.** The top-level `redis:` block only has `sentinel`, `architecture`, `auth`, `resources`; the dependency condition is `zammadConfig.redis.enabled` (`Chart.yaml:41`). Line 308 has the same short form.

Both comments sit inside `zammadConfig.redis.sentinel`, so a reader can reasonably resolve `redis.enabled` to the enclosing `zammadConfig.redis.enabled` — but since the README and the `fail` message were both just made explicit, this is now the only spot left with the ambiguous form, and that ambiguity is exactly what produced the README example bug.

```yaml
      enabled: false  # set to true to enable Redis Sentinel; with the bundled subchart
                      # (zammadConfig.redis.enabled=true) this also requires
                      # redis.sentinel.enabled=true and redis.architecture=replication
      # only used when zammadConfig.redis.enabled=false (external Redis/Sentinel); the
      # bundled subchart's sentinel service is discovered automatically and ignores this list
```

Purely cosmetic — don't hold the merge for it if you'd rather land this as-is.

Everything else checks out. I re-ran the matrix against `df720f8`: `ci/default-values.yaml` and `ci/full-values.yaml` both render, both README examples work verbatim (bundled → `rel-xyz-redis-sentinel:26379`; external → `my-sentinel-1:26379,my-sentinel-2:26379` with zero bundled-redis templates), and all three broken combinations — including the `replication` + `sentinel.enabled=false` gap — now fail fast with the correct message.

fliebe92 (dismissed):

zammad/values.yaml
mgruner: Applied — thanks for re-running the full matrix on df720f8 as well.

fliebe92 (approved):

PR #465, opened 2026-09-01
Replace deprecated minio subchart with RustFS

### LINKED ISSUE #398: Migrate away from bitnami/minio
Reason: #353 

### Ideas

- Switch to https://artifacthub.io/packages/helm/cloudpirates-minio/minio - seems to be well maintained.
- Use https://github.com/minio/operator - not sure if this is maintained.
- Drop minio subchart completely, ask users to deploy their own S3 storage or use an existing cloud service.

mgruner:
Closes #398. Last part of the move away from bitnami.

> [!IMPORTANT]
> **RustFS 1.0.0 GA shipped on 2026-09-16 and is now pinned.** The GA chart is byte-identical to `1.0.0-rc.5` in values and templates, so only the version and image tag moved. The remaining open item is the full manual migration test.

## Why

MinIO stopped publishing community Helm charts and container images, and the [MinIO operator repository was archived in March 2026](https://github.com/minio/operator). The bundled `bitnami/minio` subchart was already only usable through the `bitnamilegacy` image workaround.

[RustFS](https://rustfs.com) is S3-compatible and has an [official chart](https://github.com/rustfs/rustfs/tree/main/helm) that tracks the project's releases 1:1. The alternative `cloudpirates/rustfs` chart was evaluated and rejected: it self-describes as ALPHA and pins `rustfs:1.0.0-beta.1` from 2026-04-29 while declaring `appVersion: 1.0.0`, and its config vocabulary (`RUSTFS_LOG_LEVEL`, `RUSTFS_EXTERNAL_ADDRESS`, `RUSTFS_CORS_*`) no longer matches what the current server accepts. We only chose cloudpirates elsewhere where there was no usable official chart.

## What changed

- **Chart 19.0.0** — breaking for anyone using the optional S3 storage provider.
- `zammadConfig.minio` → `zammadConfig.rustfs`; subchart values key `minio:` → `rustfs:`.
- The subchart is pinned to single-node mode (its own default is a four-pod erasure-coded cluster), its default ingress is disabled (it would otherwise be created for `example.rustfs.com`), logs go to stdout instead of a second PVC, and the cluster default StorageClass is used.
- **Bucket provisioning.** Unlike minio's `defaultBuckets`, the rustfs chart cannot create buckets, and Zammad never creates one either — `Store::Provider::S3` only ever does `head_bucket`. A new `s3-init` init container in the zammad-init job creates it through Zammad's own S3 provider, so it cannot drift from `S3_URL`. No extra image to pin. Controlled by `zammadConfig.rustfs.bucketInitialisation` (default `true`), never applied to an external S3 service.
- README: upgrade section plus a data migration runbook. RustFS reads MinIO's on-disk layout and migrates its metadata on first start, so the attachments move as a plain **file copy between the two volumes** via a Job that mounts both — no S3 client, no data leaving the cluster, and the old volume is only read from, so it stays intact as a rollback path. Object-level copying is documented as a fallback for when both volumes cannot be attached to the same node.

## Upgrade hazard worth reviewing closely

The bitnami minio PVC carries **no** `helm.sh/resource-policy`, and the chart's default mode is `standalone` (a plain PVC, not a StatefulSet template). A naive `helm upgrade` therefore deletes the attachment volume the moment the subchart leaves the release. Step 2 of the runbook annotates the PVC with `keep` before upgrading — that step is load-bearing.

Conversely, the rustfs chart *does* set `helm.sh/resource-policy: keep`, so the new volume now survives an uninstall. That behaviour change is documented.

## Verification

- Ran a real `rustfs/rustfs:1.0.0` container with the chart's exact settings and confirmed the full path via an S3 client (path-style, SigV4, `region=zammad`): `head_bucket` → 404, `create_bucket`, `put`/`get`/`list` all succeed.
- Verified the volume-copy migration end to end, and **re-verified it against the GA image**: a MinIO volume holding four objects from 512 B to 12 MiB (covering both inline and out-of-line `part.N` storage) was copied to a pristine rustfs volume, and after adoption every object came back byte-identical with its sha256 matching the object key — which is exactly Zammad's own integrity model. Writes after the adoption work too, and `.rustfs.sys/pool.bin` is created as expected.
- Re-checked the bucket topic on GA: the server still exposes no default-bucket setting and no subcommand for it, consistent with upstream declining the feature as out of scope ([rustfs/rustfs#2179](https://github.com/rustfs/rustfs/issues/2179), closed as not planned). `s3-init` is the permanent answer here, not a stopgap.
- Confirmed `rustfs.replicaCount=0` renders a `replicas: 0` Deployment while still creating the PVC, which is what keeps the new volume pristine during phase one.
- `helm lint` and `helm template` clean on both CI value sets; `kubeconform --strict` passes (30 resources default, 40 full).
- The rendered `s3-init` script passes `bash -n` and its Ruby body passes `ruby -c`.
- Name derivation verified for a release name containing `rustfs` and for `rustfs.fullnameOverride`.

## Before merge

- [x] ~~Re-pin `zammad/Chart.yaml` to the RustFS 1.0.0 GA release~~ — done in cbbe45b.
- [x] ~~Perform a full manual migration test~~ — done, migration successful. It surfaced one gap in the runbook, fixed in a427739: the values file has to be ported to the new keys before the phase one upgrade, otherwise the render fails on the leftover `zammadConfig.minio` (intended, but surprising mid-migration) and the new subchart is not deployed at all.
- [x] ~~Enable the commented-out `rustfs` source/target blocks in `.github/updatecli.yaml`~~ — done in cbbe45b, restricted to patch level (`1.0.x`). The subchart version tracks the RustFS version exactly, so a chart minor carries a RustFS minor, and this is the store holding the attachments — the same reasoning that pins elasticsearch and postgres to patch level here.
- [x] ~~Re-check [rustfs/rustfs#7014](https://github.com/rustfs/rustfs/issues/7014)~~ — fixed upstream, and re-verified against the GA image on the re-pin. The runbook relies on this adoption behaviour, so if it regresses again the file copy silently produces an unreadable volume; that risk is why the updatecli filter above is patch-level only.

## Follow-up (separate PR, after this one is merged)

- [ ] Drop the `bitnami` entry from `.github/ct.yaml` chart-repos. It cannot go in this PR: while the target branch still carries 18.x, chart-testing needs that repo to resolve the old chart's minio dependency for the upgrade test. Only once `main` is on 19.x is it dead weight.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Replaced the bundled MinIO object-storage option with RustFS.
  - Added configurable RustFS credentials, bucket name, region, storage, logging, and security contexts.
  - Added automatic bucket provisioning during initialization, including startup retries.
  - Added support for existing RustFS secrets and external S3-compatible storage.

- **Documentation**
  - Updated configuration guidance for RustFS, OpenShift, bucket initialization, and chart upgrades.
  - Documented required value migrations, storage changes, retention behavior, and attachment migration procedures.


zammad/Chart.yaml
```diff
@@ -23,10 +23,15 @@ dependencies:
...
     condition: zammadConfig.elasticsearch.enabled
-  - name: minio
-    version: 14.10.5
-    repository: https://charts.bitnami.com/bitnami
-    condition: zammadConfig.minio.enabled
+  # Optional S3 storage provider for attachments. Replaces the former bitnami/minio
+  # subchart: MinIO stopped publishing community Helm charts and images, and the MinIO
+  # operator was archived in March 2026. RustFS is an S3-compatible drop-in replacement
+  # with an official chart. Note that it does not create buckets on its own, so the
+  # zammad-init job provisions the bucket (zammadConfig.rustfs.bucketInitialisation).
```
mgruner: This comment is not neded.

zammad/values.yaml
```diff
@@ -109,12 +109,21 @@ zammadConfig:
     host: zammad-memcached
     port: 11211
 
-  minio:
-    # enable/disable minio chart dependency
+  rustfs:
```
fliebe92, line 112: Renaming this key makes the "you must read the runbook first" hazard silent. An admin who upgrades with their existing values file keeps `zammadConfig.minio.enabled: true`, Helm ignores the now-unknown key, and:

```
helm template zam . --set zammadConfig.minio.enabled=true  →  0 occurrences of S3_URL
```

So they get no error at render time, no `S3_URL` in any pod (a `storage_provider: S3` instance then fails on every attachment), *and* the old PVC deleted, because step 2 of the runbook was never run.

This chart already `fail`s on comparable misconfigurations (`zammad.env.redisVariables`, the `storageVolume` check), and a guard here is the only part of the hazard the chart itself can enforce:

```gotemplate
{{- if or .Values.zammadConfig.minio .Values.minio -}}
{{- fail "zammadConfig.minio/minio were replaced by zammadConfig.rustfs/rustfs in chart 19.0.0. See the upgrade notes in the README - the old attachment volume needs to be protected before upgrading." -}}
{{- end -}}
```

That turns silent data loss into a message that names the new key.

zammad/README.md
```diff
@@ -191,11 +203,10 @@ memcached:
...
 
-minio:
-  podSecurityContext:
-    enabled: false
-  containerSecurityContext:
-    enabled: false
+# The rustfs subchart has no "enabled" toggle for its security contexts; emptying
+# podSecurityContext drops the hardcoded UID/GID so that OpenShift can assign its own.
+rustfs:
+  podSecurityContext: {}
```
fliebe92: This does not drop the UID/GID. Helm coalesces maps, so an empty map merges nothing and the subchart defaults survive:

```console
$ helm template zam . -f ocp.yaml --show-only charts/rustfs/templates/deployment.yaml
      securityContext:
        fsGroup: 10001
        runAsGroup: 10001
        runAsUser: 10001
```

The explicit-`null` form does work and renders `securityContext: {}`:

```yaml
rustfs:
  podSecurityContext:
    fsGroup: null
    runAsUser: null
    runAsGroup: null
```

The intro of this very section already tells the reader to [delete default keys](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key) with `null`, so the snippet currently contradicts the paragraph above it.

zammad/README.md
```diff
@@ -102,7 +102,19 @@ helm show values zammad/zammad
...
+case (`zammadConfig.rustfs.enabled: true`). RustFS is an S3-compatible object store; it replaced the
+former `minio` subchart in chart version 19.0.0, see [Upgrading](#from-chart-version-18x-to-1900).
+
+Unlike the former `minio` subchart, the `rustfs` chart does not create buckets on its own, and Zammad
+never creates the bucket either - it only reads from it. The chart therefore provisions the bucket from
+the `zammad-init` job. Set `zammadConfig.rustfs.bucketInitialisation: false` if you want to manage the
+bucket yourself. This never touches an external S3 service, only the bundled instance.
+
+The subchart is deployed as a single node with a single data volume, which is what Zammad's attachment
+store needs. If you want the redundancy of an erasure-coded cluster, switch `rustfs.mode` to
```
fliebe92: Following this literally produces a broken release. Because `values.yaml` pins `rustfs.mode.standalone.enabled: true`, setting only `mode.distributed.enabled=true` renders **both** workloads:

```console
$ helm template zam . --set zammadConfig.rustfs.enabled=true \
    --set rustfs.mode.distributed.enabled=true --set rustfs.replicaCount=4 | grep -n '^kind:\|name: zam-rustfs$'
817:kind: Deployment
819:  name: zam-rustfs
1877:kind: StatefulSet
1879:  name: zam-rustfs
```

Same name, same `rustfs.selectorLabels`, and both consume the same ConfigMap — whose `RUSTFS_VOLUMES` has by then switched to the distributed expression, so the standalone pod tries to join a cluster it is not part of. Neither `helm lint` nor `kubeconform` flags it, since a Deployment and a StatefulSet may share a name.

Two things worth adding: that `mode.standalone.enabled: false` is required as well, and that this is not an in-place change — standalone data lives in the `-data` PVC while distributed uses `volumeClaimTemplates`, so the attachments do not come along.

zammad/README.md
```diff
@@ -249,6 +260,148 @@ and `zammadConfig.cronJob.reindex.schedule` if you want to run it periodically.
...
+#    reindex cronjob, suspend it as well - it runs Zammad too and would otherwise still write.
+kubectl scale -n <namespace> deploy -l app.kubernetes.io/name=zammad --replicas=0
+kubectl patch -n <namespace> -p '{"spec":{"suspend":true}}' \
+  "$(kubectl get cronjob -n <namespace> -l app.kubernetes.io/component=zammad-cronjob-reindex -o name)"
+
+# 2. Protect the old minio volume, so that the upgrade in step 3 leaves it behind instead of
+#    deleting it. It then serves as your rollback safety net and as the source for the copy.
+kubectl annotate pvc -n <namespace> <release_name>-minio helm.sh/resource-policy=keep
+
+# 3. Phase one: upgrade with everything scaled to zero. This creates the new, empty rustfs
```
fliebe92, line 336: The runbook never tells the reader to size the new volume. `rustfs.storageclass.dataStorageSize` defaults to `10Gi` no matter how large the old minio volume was, so anyone who raised `minio.persistence.size` runs step 4 into ENOSPC — and with `backoffLimit: 0` the Job just fails with a half-copied destination.

Worth a line before this step: set `rustfs.storageclass.dataStorageSize` to at least the size of the existing `<release_name>-minio` PVC (`kubectl get pvc -n <namespace> <release_name>-minio -o jsonpath='{.spec.resources.requests.storage}'`), because it cannot be changed as easily afterwards.

zammad/templates/_helpers.tpl
```diff
@@ -210,31 +210,72 @@ redis sentinel secret name
 {{- end -}}
 {{- end -}}
 
+{{/*
+Whether the init job should provision the S3 bucket. This only applies to the
+bundled rustfs instance; an external S3 service is never modified.
+Renders a non-empty string when enabled, so it can be used in "if" conditions.
+*/}}
+{{- define "zammad.s3.initialisation" -}}
+{{- if and .Values.zammadConfig.rustfs.enabled .Values.zammadConfig.rustfs.bucketInitialisation (not .Values.zammadConfig.rustfs.externalS3Url) -}}
```
fliebe92: `externalS3Url` is not the only way `S3_URL` gets pointed elsewhere — `secrets.s3.useExisting` overrides it too (see the block at the end of `zammad.env.S3_URL`). With `zammadConfig.rustfs.enabled: true` and `secrets.s3.useExisting: true`, this renders `true`, the init container inherits the external `S3_URL` from the secret, and `s3-init` creates a bucket on that service — which is exactly what the comment above and the `values.yaml` comment promise never happens.

```gotemplate
{{- if and .Values.zammadConfig.rustfs.enabled .Values.zammadConfig.rustfs.bucketInitialisation (not .Values.zammadConfig.rustfs.externalS3Url) (not .Values.secrets.s3.useExisting) -}}
```

zammad/templates/_helpers.tpl
```diff
@@ -210,31 +210,72 @@ redis sentinel secret name
...
 - name: S3_URL
-  value: "http://$(MINIO_ROOT_USER):$(MINIO_ROOT_PASSWORD)@{{ include "zammad.fullname" . }}-minio:9000/zammad?region=zammad&force_path_style=true"
+  value: "http://$(RUSTFS_ACCESS_KEY):$(RUSTFS_SECRET_KEY)@{{ include "zammad.rustfsEndpoint" . }}/{{ $bucket }}?region={{ $region }}&force_path_style=true"
 {{- else -}}
 - name: S3_URL
-  value: "http://{{ .Values.minio.auth.rootUser }}:{{ .Values.minio.auth.rootPassword }}@{{ include "zammad.fullname" . }}-minio:9000/zammad?region=zammad&force_path_style=true"
+  value: "http://{{ .Values.rustfs.secret.rustfs.access_key }}:{{ .Values.rustfs.secret.rustfs.secret_key }}@{{ include "zammad.rustfsEndpoint" . }}/{{ $bucket }}?region={{ $region }}&force_path_style=true"
 {{- end -}}
 {{- end -}}
 {{- if .Values.secrets.s3.useExisting -}}
```
fliebe92: Not introduced here — `main` behaves identically — but since the helper is being rewritten anyway: the `{{-` here eats the newline after the preceding `value:` line, so combining the bundled instance with `secrets.s3.useExisting` renders invalid YAML and `helm template` fails outright.

```console
$ helm template zam . --set zammadConfig.rustfs.enabled=true --set secrets.s3.useExisting=true \
    --set secrets.s3.secretName=my-s3 --set secrets.s3.secretKey=url
Error: YAML parse error on zammad/templates/cronjob-reindex.yaml: error converting YAML to JSON: yaml: line 113: block sequence entries are not allowed in this context

# the offending line:
  value: "http://zammadadmin:zammadadmin@zam-rustfs-svc:9000/zammad?…"- name: S3_URL
```

`{{ if .Values.secrets.s3.useExisting -}}` fixes the render. Even then the container ends up with two `S3_URL` entries (last one wins), so an `else` against the bundled branch would be the cleaner shape — your call whether that belongs in this PR.

zammad/templates/configmap-init.yaml
```diff
@@ -72,3 +72,41 @@ data:
...
+      client = Store::Provider::S3.client
+
+      attempts = 0
+
+      begin
+        client.head_bucket(bucket: bucket)
+        puts "S3 bucket '#{bucket}' already exists."
+      rescue Aws::S3::Errors::NotFound, Aws::S3::Errors::NoSuchBucket
+        puts "S3 bucket '#{bucket}' does not exist yet, creating it…"
+        client.create_bucket(bucket: bucket)
```
fliebe92: Nit, both self-heal via `restartPolicy: OnFailure` so this is only about not scaring people with a crashed init container in the logs:

- `create_bucket` is not guarded against `Aws::S3::Errors::BucketAlreadyOwnedByYou` / `BucketAlreadyExists`. Step 5 of the runbook is the realistic window: the init job and the rustfs pod come up together, and a `head_bucket` that hits the server mid-adoption can 404 on a bucket that exists a moment later.
- A `Seahorse::Client::NetworkingError` raised *inside* this `rescue` clause is not caught by the sibling `rescue` below — Ruby does not re-enter rescue clauses of the same `begin`. Wrapping just the `create_bucket` call, or moving the networking rescue to an outer `begin`, covers it.

zammad/README.md
```diff
@@ -249,6 +260,148 @@ and `zammadConfig.cronJob.reindex.schedule` if you want to run it periodically.
...
+  backoffLimit: 0
+  template:
+    spec:
+      restartPolicy: Never
+      securityContext:
+        runAsUser: 0
+      containers:
+        - name: copy
+          image: alpine:3
+          command: ["sh", "-ec", "cp -a /src/. /dst/ && chown -R 10001:10001 /dst && du -sh /dst"]
```
fliebe92: Two nits on the copy Job:

- On ext4-backed PVCs `cp -a /src/.` also copies `lost+found`, which lands in rustfs's data root next to the buckets. Worth excluding, or at least mentioning.
- `runAsUser: 0` above is rejected outright by a namespace enforcing the `restricted` Pod Security Standard, which is worth a note given the chart documents OpenShift support. Setting `fsGroup` on the Job instead of `chown -R` would avoid needing root at all, though it costs a relabel of the whole volume on mount.

zammad/Chart.yaml
mgruner: Removed in 89ef167.

zammad/values.yaml
mgruner: Confirmed, and it is as quiet as you describe: `helm template --set zammadConfig.minio.enabled=true` rendered successfully with zero `S3_URL` occurrences.

Guard added in 89ef167. It lives inline at the top of `zammad.env.S3_URL` rather than in its own helper, since that is the only place it is needed, and it fires both for `zammadConfig.minio` and for a leftover `minio:` subchart block.

zammad/README.md
mgruner: You are right, and thanks for the exact repro — I reproduced both halves:

```console
# podSecurityContext: {}
      securityContext:
        fsGroup: 10001
        runAsGroup: 10001
        runAsUser: 10001

# explicit nulls
      securityContext:
        {}
```

Switched the snippet to the null form in 89ef167, so it now matches the paragraph above it.

zammad/README.md
mgruner: Reproduced — a `Deployment` and a `StatefulSet` both named `<release>-rustfs`, and neither `helm lint` nor `kubeconform` objects.

Rather than document the extra flag, we removed the whole passage in 89ef167. The paragraph was the only thing inviting readers down that path, and configuring a multi-node topology belongs in the subchart's own documentation rather than here. That leaves no incorrect instruction behind, which was the actual defect.

zammad/README.md
mgruner: Good catch. Added before step 3 in 89ef167, naming the key and — importantly — where it has to go:

> Set `rustfs.storageclass.dataStorageSize` in `my-values.yaml` to at least the size of the old volume beforehand - it defaults to 10Gi and a PVC cannot be shrunk again later.

The "in `my-values.yaml`" part matters: passing it via `--set` in step 3 would be dropped by the `--reset-values` in step 5, Helm would patch the PVC back to 10Gi, and Kubernetes rejects shrinking a PVC — so the second upgrade would fail with the data already copied.

zammad/templates/_helpers.tpl
mgruner: Correct, `(not .Values.secrets.s3.useExisting)` added in 89ef167. Verified that `s3-init` no longer renders in that combination and still renders for the bundled instance.

zammad/templates/_helpers.tpl
mgruner: Confirmed, including the parse error verbatim. Fixed in 89ef167, and we took the `else` shape you suggested: the three sources are now mutually exclusive branches, with `secrets.s3.useExisting` first so that the value which used to win (last one rendered) still wins.

All three configurations now render exactly one `S3_URL`.

zammad/templates/configmap-init.yaml
mgruner: Both correct. I checked the Ruby semantics with a standalone snippet and the exception does escape uncaught, so the retry never applied to anything raised while creating the bucket.

89ef167 wraps the `create_bucket` call and rescues `BucketAlreadyOwnedByYou` / `BucketAlreadyExists`, and moves the networking rescue to an outer `begin` so `retry` covers the whole sequence.

zammad/README.md
mgruner: Both addressed in 89ef167.

`lost+found` is now dropped rather than excluded, which keeps the command a one-liner: `cp -a /src/. /dst/ && rm -rf /dst/lost+found && chown -R 10001:10001 /dst`. Verified that dotfiles including `.minio.sys` still come across.

And there is now a note that the Job's root requirement is rejected under the `restricted` Pod Security Standard, pointing at `fsGroup` instead of `chown` as the alternative there.

zammad/templates/_helpers.tpl
```diff
@@ -211,38 +211,83 @@ redis sentinel secret name
...
+S3 access URL
+*/}}
+{{- define "zammad.env.S3_URL" -}}
+{{/* Helm silently ignores unknown keys, so without this an upgrade with an unchanged
+     values.yaml would leave Zammad without an S3_URL and delete the old attachment
+     volume unnoticed. */}}
+{{- if or .Values.zammadConfig.minio .Values.minio -}}
+{{- fail "zammadConfig.minio/minio were replaced by zammadConfig.rustfs/rustfs in chart 19.0.0. See the upgrade notes in the README - the old attachment volume needs to be protected before upgrading." -}}
+{{- end -}}
 {{- if .Values.secrets.s3.useExisting -}}
```
fliebe92: Thanks for taking the `else` shape — but the ordering does change behaviour, contrary to "preserving which value won before". That holds for the bundled case, not for `externalS3Url` + `useExisting`: in 18.x the `useExisting` block sat *inside* the `{{- with externalS3Url }}…{{- else }}` branch, so it never rendered at all when `externalS3Url` was set.

```console
# main (18.x)
$ helm template zam . --set zammadConfig.minio.externalS3Url=https://u:p@ext/bucket \
    --set secrets.s3.useExisting=true --set secrets.s3.secretName=my-s3 --set secrets.s3.secretKey=url
- name: S3_URL
  value: "https://u:p@ext/bucket"      ← externalS3Url won

# 89ef167
- name: S3_URL
  valueFrom:
    secretKeyRef:
      name: my-s3                       ← secrets.s3 now wins
```

The new order is arguably the better one — the secret is the more explicit choice and keeps credentials out of the values file — but for anyone carrying both keys this is a silent switch to a different endpoint on upgrade, and it is not in the 19.0.0 notes. Either reorder to `externalS3Url` → `useExisting` → bundled, or keep it and add a line to the upgrade section.

zammad/templates/_helpers.tpl
```diff
@@ -211,38 +211,83 @@ redis sentinel secret name
...
+
+{{/*
+S3 access URL
+*/}}
+{{- define "zammad.env.S3_URL" -}}
+{{/* Helm silently ignores unknown keys, so without this an upgrade with an unchanged
+     values.yaml would leave Zammad without an S3_URL and delete the old attachment
+     volume unnoticed. */}}
+{{- if or .Values.zammadConfig.minio .Values.minio -}}
+{{- fail "zammadConfig.minio/minio were replaced by zammadConfig.rustfs/rustfs in chart 19.0.0. See the upgrade notes in the README - the old attachment volume needs to be protected before upgrading." -}}
```
fliebe92: The guard works, verified both halves (fires on `zammadConfig.minio` and on a leftover `minio:` block; a clean values file still renders). Two small things:

- It also fires for someone who merely kept `zammadConfig.minio.enabled: false` — the old default — in their values file and never used the S3 provider at all. `.Values.zammadConfig.minio` is a non-empty dict either way. They then get a hard upgrade failure telling them "the old attachment volume needs to be protected", which does not apply to them. Keying the volume half of the sentence on `.Values.zammadConfig.minio.enabled` would keep the message accurate for both groups.
- Helm attributes the failure to wherever the helper happens to be evaluated first, which here is `execution error at (zammad/templates/tests/run-tests.yaml:61:18)`. Harmless, but it does send people to a file that has nothing to do with the problem — worth knowing when someone reports it.

zammad/README.md
```diff
@@ -249,6 +259,160 @@ and `zammadConfig.cronJob.reindex.schedule` if you want to run it periodically.
...
+            - { name: src, mountPath: /src }
+            - { name: dst, mountPath: /dst }
+      volumes:
+        - name: src
+          persistentVolumeClaim: { claimName: <release_name>-minio }
+        - name: dst
+          persistentVolumeClaim: { claimName: <release_name>-rustfs-data }
+EOF
+kubectl wait -n <namespace> --for=condition=complete --timeout=2h job/minio-to-rustfs
+kubectl logs -n <namespace> job/minio-to-rustfs
```
fliebe92, line 382: Collateral from the `lost+found` fix: `du -sh /dst` was dropped from the Job command, but this `kubectl logs` line survived. It now prints nothing, and with it goes the one sanity check that showed the copy actually landed something. Appending `&& du -sh /dst` again keeps the step meaningful.

zammad/README.md
```diff
@@ -249,6 +259,160 @@ and `zammadConfig.cronJob.reindex.schedule` if you want to run it periodically.
...
+# 6. Once you have verified that attachments open correctly in Zammad, clean up at your
+#    convenience. Until then the old volume is untouched and you can roll back to it.
+kubectl delete job -n <namespace> minio-to-rustfs
+kubectl delete pvc -n <namespace> <release_name>-minio
+```
+
+The Job runs as root for the `chown`, which a namespace enforcing the `restricted`
+[Pod Security Standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/) (or
+OpenShift's default SCC) rejects. There, drop `securityContext.runAsUser` and the `chown` line and
+set `securityContext.fsGroup` to the UID rustfs runs as instead, which makes the kubelet adjust
```
fliebe92: `fsGroup` is a GID, not a UID. And in the OpenShift case this paragraph is written for, neither is knowable in advance: the SCC assigns the range per namespace, which is the whole reason the OpenShift snippet above nulls out `runAsUser`/`fsGroup`. Leaving `fsGroup` unset is the right advice there — the SCC injects it, and the kubelet relabels the volume to match the same GID the rustfs pod gets. Only a cluster where you pin the IDs yourself needs an explicit value.

zammad/templates/_helpers.tpl
mgruner: You are right, and my "preserving which value won before" was simply wrong for that combination — thanks for pinning it down. Reproduced against `main`:

```console
# main (18.x): externalS3Url wins
- name: S3_URL
  value: "https://u:p@ext/bucket"

# 89ef167: the secret won
      secretKeyRef:
        name: my-s3
```

Reordered to `externalS3Url` → `useExisting` → bundled in da4a558, so 18.x behaviour is restored and no upgrade note is needed. I took the conservative option deliberately: a silent endpoint switch is the same class of surprise the new removed-keys guard exists to prevent. There is now a comment on the branch recording why the order matters, so it does not get "tidied up" later.

All four configurations still render exactly one `S3_URL`.

zammad/templates/_helpers.tpl
mgruner: Good point on the false positive — that hits everyone who copied the full `values.yaml`, which is a lot of people.

da4a558 keeps the failure in both cases, since the key has to be renamed either way, but the volume sentence is now conditional on `zammadConfig.minio.enabled`:

```console
# enabled: false, or a leftover minio: block
zammadConfig.minio/minio were replaced by zammadConfig.rustfs/rustfs in chart 19.0.0.

# enabled: true
... in chart 19.0.0. The old attachment volume needs to be protected before upgrading, see the data migration in the upgrade notes.
```

On the attribution: agreed, and as far as I can tell it is not fixable — Helm reports whichever template evaluated the helper first, and any always-rendered location has the same problem. Noted here for whoever gets the bug report.

zammad/README.md
mgruner: Correct, that was collateral. Re-appended in da4a558, still a one-liner:

```
cp -a /src/. /dst/ && rm -rf /dst/lost+found && chown -R 10001:10001 /dst && du -sh /dst
```

zammad/README.md
mgruner: Both wrong on my side — GID, and unknowable in advance in exactly the environment the paragraph is written for. Rewritten in da4a558 to say what you describe: drop `runAsUser` and the `chown`, leave `fsGroup` unset so the SCC assigns it and the kubelet relabels the volume to the same GID, and pin a value only when you assign the IDs yourself.

fliebe92 (dismissed):

fliebe92 (dismissed):

fliebe92 (dismissed):

.github/updatecli.yaml
```diff
@@ -33,14 +33,14 @@ sources:
...
-  #     versionfilter:
-  #       pattern: "14.x.x"
+  rustfs:
+    name: rustfs
+    kind: helmchart
+    spec:
+      url: https://charts.rustfs.com
+      name: rustfs
+      versionfilter:
+        pattern: "1.x.x"
```
fliebe92, line 43: `1f57bd1` widens this from `1.0.x` to `1.x.x` and deletes the rationale comment along with it. That makes rustfs an outlier against the convention the rest of this file follows: every subchart whose version bump can carry a data-layer change gets a `Patch level only:` comment plus a pinned minor — `eck-elasticsearch` at `0.20.x`, `postgres` at `0.19.x`. The only two on `0.x.x` are `memcached` and `redis`, and both are caches where losing the contents costs nothing.

rustfs is now the one stateful subchart on unrestricted minor tracking, and it holds the attachments. The chart version tracks the RustFS server 1:1 — `image.rustfs.tag` is empty, so it resolves to `.Chart.AppVersion`, and the render confirms it:

```console
$ helm template zam . --set zammadConfig.rustfs.enabled=true | grep 'image: "rustfs'
          image: "rustfs/rustfs:1.0.0"
```

So chart `1.1.0` would put a RustFS *server* minor onto an existing attachment volume, and `versionincrement: patch` below ships that as a Zammad chart *patch* release.

To be fair to the change: the rustfs chart is on `1.x`, where semver says minors are compatible, unlike the `0.x` cloudpirates charts where a minor is a breaking release. `1.x.x` is the same risk class as `alpine` `3.x.x`, so this is a judgement call rather than a defect.

The ask is just to write the judgement down. The deleted comment cited rustfs/rustfs#7014, and the PR description still notes that a regression there silently produces an unreadable volume. Without a replacement note, the next person either sees an unexplained outlier and re-pins it, or merges an auto-bump PR without the data-layer check the other stateful subcharts deliberately get.

mgruner: There are two patch-level only filters in the updatecli: postgres and es. They are both documented. Others are more liberal, and not documented, so this fits in well. The regression during the beta period is not a good argument for requiring a tighter pinning. I'd like to try it like it is.

fliebe92 (approved):

zammad/templates/_helpers.tpl
```diff
@@ -210,39 +210,91 @@ redis sentinel secret name
...
+- name: RUSTFS_SECRET_KEY
   valueFrom:
     secretKeyRef:
-      key: root-password
-      name: {{ .Values.minio.auth.existingSecret }}
+      key: RUSTFS_SECRET_KEY
+      name: {{ .Values.rustfs.secret.existingSecret }}
 - name: S3_URL
-  value: "http://$(MINIO_ROOT_USER):$(MINIO_ROOT_PASSWORD)@{{ include "zammad.fullname" . }}-minio:9000/zammad?region=zammad&force_path_style=true"
+  value: "http://$(RUSTFS_ACCESS_KEY):$(RUSTFS_SECRET_KEY)@{{ include "zammad.rustfsEndpoint" . }}/{{ $bucket }}?region={{ $region }}&force_path_style=true"
```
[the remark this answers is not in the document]
mgruner: That's up to the admin and the same behaviour like with minio before.

zammad/values.yaml
```diff
@@ -787,37 +813,50 @@ memcached:
...
+      enabled: true
+    distributed:
+      enabled: false
+  replicaCount: 1
 
-  defaultBuckets: zammad
+  secret:
+    rustfs:
+      access_key: zammadadmin
+      secret_key: zammadadmin
```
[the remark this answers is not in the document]
mgruner: That's up to the admin and the same behaviour like with minio before.

PR #470, opened 2026-09-10
Exclude the Elasticsearch host from proxy routing on new installations

mgruner:
Connections between pods of the same cluster are not external and must not be routed through a configured proxy. On a new installation `proxy_no` is now seeded with the Elasticsearch host, so that Zammad's own HTTP client contacts Elasticsearch directly.

This is the Helm counterpart of zammad/zammad!14238, which fixes the same issue in `bin/docker-entrypoint`. The chart does not use that entrypoint — it mounts its own init scripts from `configmap-init.yaml` — so the change has to be made here separately.

The call sits in the first-install branch of `postgresql-init`, not in `elasticsearch-init`. The latter is the closer analogue of the entrypoint's `es config` block, but by the time it runs, `postgresql-init` has already executed `db:seed`, which makes `User.any?` true — a first install is no longer detectable there. Placing it in the first-install branch also means an admin's own No Proxy value is never overwritten on a later upgrade.

It is skipped when `zammadConfig.elasticsearch.initialisation` is `false`, the chart's equivalent of the entrypoint's `ELASTICSEARCH_ENABLED` skip. The host comes from the existing `zammad.elasticsearchHost` helper, so it is correct both for the bundled ECK Elasticsearch and for an external one.

Existing installations are deliberately out of scope, per the parent story: no migration is written, and affected admins set the field by hand, which already works today.

Issue: https://github.com/zammad/coordination-scrum/issues/1735
Parent story: https://github.com/zammad/zammad/issues/6228

## Testing

Manually verified with `helm install`: a fresh installation comes up with the Elasticsearch host in the No Proxy field. Rendering was checked for the bundled ECK Elasticsearch, for an external host, and for `initialisation: false` (where the block is absent); `helm lint` is clean.

There is no CI assertion for the seeded value. One was written and then dropped. It never found a defect in this change, but diagnosing its own failures surfaced two unrelated properties: `helm install --wait` does not wait for Jobs (`--wait-for-jobs` is a separate flag), so `helm test` ran against half-initialised releases; and `Setting.get` serves an in-process cache for `ZAMMAD_SETTING_TTL` (15s), so a test pod that booted before the write kept reading the stale value. Both are written up in the comments below. The block stays render-checked only, as the external-Elasticsearch path already was.

The `--wait-for-jobs` finding outlived the test and is documented in the README, since it applies to anyone scripting installs or using ArgoCD.

Two changes to the helm test job are kept from that work: the test container no longer inherits railsserver's probes, which target a port the rake pod never listens on, and `activeDeadlineSeconds` goes from 30 to 600 to leave headroom for a slow Rails boot.

## Summary by CodeRabbit

- **Bug Fixes**
  - Elasticsearch hosts are now excluded from proxy routing during initial PostgreSQL setup only when Elasticsearch initialization is enabled.

- **Chores**
  - Updated the Helm chart version to 18.1.0.

### DIFF
--- a/zammad/templates/configmap-init.yaml
+++ b/zammad/templates/configmap-init.yaml
@@ -14,7 +14,7 @@ data:
 
     if ! (bundle exec rails r 'puts User.any?' 2> /dev/null | grep -q true); then
         bundle exec rake db:migrate db:seed
-    {{- if .Values.zammadConfig.elasticsearch.enabled }}
+    {{- if .Values.zammadConfig.elasticsearch.initialisation }}
 
         # Connections between pods of the same cluster are not external, so exclude Elasticsearch
         #   from proxy routing. Only done on first install, so that admins can change the value later.
--- a/zammad/templates/configmap-init.yaml
+++ b/zammad/templates/configmap-init.yaml
@@ -30,6 +30,9 @@ data:
         bundle exec rails r "Locale.sync; Translation.sync"
     fi
 
+    # Also store a value in the Rails cache that can be used to detect job completion.
+    bundle exec rails r "Rails.cache.write('helm-postgresql-init-done-{{ .Release.Revision }}', true, { expires_in: 1.hour })"
+
     echo "postgresql init complete :)"
   postgresql-init-post: |-
     #!/bin/bash
--- a/zammad/templates/configmap-init.yaml
+++ b/zammad/templates/configmap-init.yaml
@@ -14,11 +14,11 @@ data:
 
     if ! (bundle exec rails r 'puts User.any?' 2> /dev/null | grep -q true); then
         bundle exec rake db:migrate db:seed
-    {{- if .Values.zammadConfig.elasticsearch.initialisation }}
+    {{- if .Values.zammadConfig.elasticsearch.enabled }}
 
         # Connections between pods of the same cluster are not external, so exclude Elasticsearch
         #   from proxy routing. Only done on first install, so that admins can change the value later.
-        ELASTICSEARCH_HOST={{ include "zammad.elasticsearchHost" . }}
+        ELASTICSEARCH_HOST="{{ include "zammad.elasticsearchHost" . }}"
         echo "Excluding ${ELASTICSEARCH_HOST} from proxy routing..."
         bundle exec rails r "Setting.set('proxy_no', '${ELASTICSEARCH_HOST}')"
     {{- end }}
--- a/zammad/templates/configmap-init.yaml
+++ b/zammad/templates/configmap-init.yaml
@@ -16,8 +16,8 @@ data:
         bundle exec rake db:migrate db:seed
     {{- if .Values.zammadConfig.elasticsearch.initialisation }}
 
-        # Connections between pods of the same cluster are not external, so exclude Elasticsearch
-        #   from proxy routing. Only done on first install, so that admins can change the value later.
+        # Elasticsearch is contacted directly, not through a configured proxy. Only done on first
+        #   install, so that admins can change the value later.
         ELASTICSEARCH_HOST="{{ include "zammad.elasticsearchHost" . }}"
         echo "Excluding ${ELASTICSEARCH_HOST} from proxy routing..."
         bundle exec rails r "Setting.set('proxy_no', '${ELASTICSEARCH_HOST}')"
--- a/zammad/templates/configmap-init.yaml
+++ b/zammad/templates/configmap-init.yaml
@@ -14,6 +14,14 @@ data:
 
     if ! (bundle exec rails r 'puts User.any?' 2> /dev/null | grep -q true); then
         bundle exec rake db:migrate db:seed
+    {{- if .Values.zammadConfig.elasticsearch.initialisation }}
+
+        # Connections between pods of the same cluster are not external, so exclude Elasticsearch
+        #   from proxy routing. Only done on first install, so that admins can change the value later.
+        ELASTICSEARCH_HOST={{ include "zammad.elasticsearchHost" . }}
+        echo "Excluding ${ELASTICSEARCH_HOST} from proxy routing..."
+        bundle exec rails r "Setting.set('proxy_no', '${ELASTICSEARCH_HOST}')"
+    {{- end }}
     else
         echo Executing migrations...
         bundle exec rake db:migrate
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -31,6 +31,26 @@ data:
           {{- end }}
 
           {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
+          # `helm install --wait` does not wait for Jobs (that needs --wait-for-jobs) and the init
+          #   job is deliberately not a Helm hook (see job-init.yaml), so this test can start while
+          #   postgresql-init is still migrating and seeding. elasticsearch-init sets es_url and only
+          #   runs once postgresql-init succeeded, so a present es_url means proxy_no has been
+          #   written if it was going to be. Observed waits are up to 90s.
+          # Setting.get caches for ZAMMAD_SETTING_TTL (15s), so force a reload on every attempt.
+          puts 'Waiting for the database initialisation to finish…'
+          started  = Time.current
+          deadline = 5.minutes.from_now
+          while Time.current < deadline
+            Setting.reload
+            break if Setting.get('es_url').present?
+
+            sleep 5
+          end
+          if Setting.get('es_url').blank?
+            raise 'Timed out waiting for the database initialisation to finish.'
+          end
+          puts "  Database initialisation has finished, waited #{(Time.current - started).round}s."
+
           puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
           # Seed counts tell apart "db:seed never finished" from "the write itself did not happen":
           #   users appear with the 2nd seed file, overviews and schedulers near the end.
@@ -56,7 +76,9 @@ metadata:
   annotations:
     "helm.sh/hook": test
 spec:
-  activeDeadlineSeconds: 30
+  # Outlasts the wait for the database initialisation above, so that its own timeout message
+  #   reaches the log instead of Kubernetes killing the pod with DeadlineExceeded first.
+  activeDeadlineSeconds: 600
   backoffLimit: 0
   template:
     metadata:
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -32,12 +32,15 @@ data:
 
           {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
           puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
+          # Seed counts tell apart "db:seed never finished" from "the write itself did not happen":
+          #   users appear with the 2nd seed file, overviews and schedulers near the end.
+          state = "proxy_no=#{Setting.get('proxy_no').inspect}, system_init_done=#{Setting.get('system_init_done').inspect}, users=#{User.count}, overviews=#{Overview.count}, schedulers=#{Scheduler.count}"
           # Mirrors UserAgent::HttpClient#proxy_no: entries are comma separated and matched exactly.
           proxy_no = Setting.get('proxy_no').to_s.split(',').map { |entry| entry.strip.downcase }
           if proxy_no.exclude?('{{ include "zammad.elasticsearchHost" . | lower }}')
-            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . | lower }}' is missing from the 'proxy_no' setting, which is #{Setting.get('proxy_no').inspect}. Expected on a fresh install; benign if this release was installed onto an already-seeded database, or if the value was changed by an admin."
+            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . | lower }}' is missing from the 'proxy_no' setting (#{state})."
           end
-          puts '  Elasticsearch host is excluded from proxy routing.'
+          puts "  Elasticsearch host is excluded from proxy routing (#{state})."
           {{- end }}
 
         end
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -13,6 +13,20 @@ data:
         desc 'Runs a set of Helm tests'
         task test: :environment do |_task, args|
 
+          {{- if .Values.zammadConfig.initJob.enabled }}
+          # Helm does not wait for jobs, so make sure zammad-init completed.
+          puts 'Waiting for the database initialisation to finish…'
+          (started, deadline)  = Time.current, 5.minutes.from_now
+          while Time.current < deadline
+            break if Rails.cache.read('helm-postgresql-init-done-{{ .Release.Revision }}')
+            sleep 5
+          end
+          if !Rails.cache.read('helm-postgresql-init-done-{{ .Release.Revision }}')
+            raise 'Timed out waiting for the database initialisation to finish.'
+          end
+          puts "  Database initialisation has finished, waited #{(Time.current - started).round}s."
+          {{- end }}
+
           puts 'Checking if a Ruby Tempfile can be created…'
           Tempfile.create do |_f|
             puts '  Tempfile file was created successfully.'
@@ -31,36 +45,12 @@ data:
           {{- end }}
 
           {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
-          # `helm install --wait` does not wait for Jobs (that needs --wait-for-jobs) and the init
-          #   job is deliberately not a Helm hook (see job-init.yaml), so this test can start while
-          #   postgresql-init is still migrating and seeding. elasticsearch-init sets es_url and only
-          #   runs once postgresql-init succeeded, so a present es_url means proxy_no has been
-          #   written if it was going to be. Observed waits are up to 90s.
-          # Setting.get caches for ZAMMAD_SETTING_TTL (15s), so force a reload on every attempt.
-          puts 'Waiting for the database initialisation to finish…'
-          started  = Time.current
-          deadline = 5.minutes.from_now
-          while Time.current < deadline
-            Setting.reload
-            break if Setting.get('es_url').present?
-
-            sleep 5
-          end
-          if Setting.get('es_url').blank?
-            raise 'Timed out waiting for the database initialisation to finish.'
-          end
-          puts "  Database initialisation has finished, waited #{(Time.current - started).round}s."
-
           puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
-          # Seed counts tell apart "db:seed never finished" from "the write itself did not happen":
-          #   users appear with the 2nd seed file, overviews and schedulers near the end.
-          state = "proxy_no=#{Setting.get('proxy_no').inspect}, system_init_done=#{Setting.get('system_init_done').inspect}, users=#{User.count}, overviews=#{Overview.count}, schedulers=#{Scheduler.count}"
-          # Mirrors UserAgent::HttpClient#proxy_no: entries are comma separated and matched exactly.
-          proxy_no = Setting.get('proxy_no').to_s.split(',').map { |entry| entry.strip.downcase }
-          if proxy_no.exclude?('{{ include "zammad.elasticsearchHost" . | lower }}')
-            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . | lower }}' is missing from the 'proxy_no' setting (#{state}). Expected on a fresh install; benign if this release was installed onto an already-seeded database, or if the value was changed by an admin."
+          proxy_no = Setting.get('proxy_no')
+          if proxy_no != '{{ include "zammad.elasticsearchHost" . }}'
+            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . }}' is missing from the 'proxy_no' setting (#{proxy_no})."
           end
-          puts "  Elasticsearch host is excluded from proxy routing (#{state})."
+          puts "  Elasticsearch host is excluded from proxy routing (#{proxy_no})."
           {{- end }}
 
         end
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -21,6 +21,7 @@ data:
           #   initialisation is done and any check below can rely on its result.
           # Setting.get caches for ZAMMAD_SETTING_TTL (15s), so force a reload on every attempt.
           puts 'Waiting for the database initialisation to finish…'
+          started  = Time.current
           deadline = 8.minutes.from_now
           while Time.current < deadline
             Setting.reload
@@ -31,7 +32,7 @@ data:
           if Setting.get('es_url').blank?
             raise 'Timed out waiting for the database initialisation to finish.'
           end
-          puts '  Database initialisation has finished.'
+          puts "  Database initialisation has finished, waited #{(Time.current - started).round}s."
           {{- end }}
 
           puts 'Checking if a Ruby Tempfile can be created…'
@@ -88,8 +89,10 @@ spec:
         {{- include "zammad.podSpec.deployment" . | nindent 8 }}
         containers:
           - name: zammad-run-tests
-            # Use securityContext etc. from railsserver to have identical setup.
-            {{- include "zammad.containerSpec" (merge (dict "containerConfig" .Values.zammadConfig.railsserver) .) | nindent 12 }}
+            # Use securityContext etc. from railsserver to have identical setup, but not its
+            #   probes: this container runs a rake task and never listens on the railsserver port,
+            #   so the startup probe would always fail and kill it after failureThreshold periods.
+            {{- include "zammad.containerSpec" (merge (dict "containerConfig" (omit .Values.zammadConfig.railsserver "startupProbe" "livenessProbe" "readinessProbe")) .) | nindent 12 }}
             command: ['bundle']
             args: ['exec', 'rake', 'zammad:helm:test']
             env:
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -46,6 +46,7 @@ data:
 
           {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
           puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
+          Setting.reload # Settings cache might be stale, so refresh.
           proxy_no = Setting.get('proxy_no')
           if proxy_no != '{{ include "zammad.elasticsearchHost" . }}'
             raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . }}' is missing from the 'proxy_no' setting (#{proxy_no})."
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -35,7 +35,7 @@ data:
           # Mirrors UserAgent::HttpClient#proxy_no: entries are comma separated and matched exactly.
           proxy_no = Setting.get('proxy_no').to_s.split(',').map { |entry| entry.strip.downcase }
           if proxy_no.exclude?('{{ include "zammad.elasticsearchHost" . | lower }}')
-            raise "Elasticsearch host is missing from the 'proxy_no' setting. Expected on a fresh install; benign if this release was installed onto an already-seeded database, or if the value was changed by an admin."
+            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . | lower }}' is missing from the 'proxy_no' setting, which is #{Setting.get('proxy_no').inspect}. Expected on a fresh install; benign if this release was installed onto an already-seeded database, or if the value was changed by an admin."
           end
           puts '  Elasticsearch host is excluded from proxy routing.'
           {{- end }}
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -13,6 +13,27 @@ data:
         desc 'Runs a set of Helm tests'
         task test: :environment do |_task, args|
 
+          {{- if .Values.zammadConfig.elasticsearch.initialisation }}
+          # `helm install --wait` returns before the init job completes, because the job is
+          #   deliberately not a Helm hook (see job-init.yaml), so this test can start while
+          #   postgresql-init is still migrating and seeding. elasticsearch-init sets es_url and
+          #   only runs once postgresql-init succeeded, so a present es_url means the database
+          #   initialisation is done and any check below can rely on its result.
+          # Setting.get caches for ZAMMAD_SETTING_TTL (15s), so force a reload on every attempt.
+          puts 'Waiting for the database initialisation to finish…'
+          deadline = 8.minutes.from_now
+          while Time.current < deadline
+            Setting.reload
+            break if Setting.get('es_url').present?
+
+            sleep 5
+          end
+          if Setting.get('es_url').blank?
+            raise 'Timed out waiting for the database initialisation to finish.'
+          end
+          puts '  Database initialisation has finished.'
+          {{- end }}
+
           puts 'Checking if a Ruby Tempfile can be created…'
           Tempfile.create do |_f|
             puts '  Tempfile file was created successfully.'
@@ -56,7 +77,8 @@ metadata:
   annotations:
     "helm.sh/hook": test
 spec:
-  activeDeadlineSeconds: 30
+  # Long enough to outlast the wait for the database initialisation above.
+  activeDeadlineSeconds: 600
   backoffLimit: 0
   template:
     metadata:
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -32,8 +32,10 @@ data:
 
           {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
           puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
-          if Setting.get('proxy_no').to_s.exclude?('{{ include "zammad.elasticsearchHost" . }}')
-            raise "Elasticsearch host is missing from the 'proxy_no' setting."
+          # Mirrors UserAgent::HttpClient#proxy_no: entries are comma separated and matched exactly.
+          proxy_no = Setting.get('proxy_no').to_s.split(',').map { |entry| entry.strip.downcase }
+          if proxy_no.exclude?('{{ include "zammad.elasticsearchHost" . | lower }}')
+            raise "Elasticsearch host is missing from the 'proxy_no' setting. Expected on a fresh install; benign if this release was installed onto an already-seeded database, or if the value was changed by an admin."
           end
           puts '  Elasticsearch host is excluded from proxy routing.'
           {{- end }}
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -30,6 +30,14 @@ data:
           puts 'Storage volume not enabled, not testing it…'
           {{- end }}
 
+          {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
+          puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
+          if Setting.get('proxy_no').to_s.exclude?('{{ include "zammad.elasticsearchHost" . }}')
+            raise "Elasticsearch host is missing from the 'proxy_no' setting."
+          end
+          puts '  Elasticsearch host is excluded from proxy routing.'
+          {{- end }}
+
         end
       end
     end
--- a/zammad/templates/tests/run-tests.yaml
+++ b/zammad/templates/tests/run-tests.yaml
@@ -13,28 +13,6 @@ data:
         desc 'Runs a set of Helm tests'
         task test: :environment do |_task, args|
 
-          {{- if .Values.zammadConfig.elasticsearch.initialisation }}
-          # `helm install --wait` returns before the init job completes, because the job is
-          #   deliberately not a Helm hook (see job-init.yaml), so this test can start while
-          #   postgresql-init is still migrating and seeding. elasticsearch-init sets es_url and
-          #   only runs once postgresql-init succeeded, so a present es_url means the database
-          #   initialisation is done and any check below can rely on its result.
-          # Setting.get caches for ZAMMAD_SETTING_TTL (15s), so force a reload on every attempt.
-          puts 'Waiting for the database initialisation to finish…'
-          started  = Time.current
-          deadline = 8.minutes.from_now
-          while Time.current < deadline
-            Setting.reload
-            break if Setting.get('es_url').present?
-
-            sleep 5
-          end
-          if Setting.get('es_url').blank?
-            raise 'Timed out waiting for the database initialisation to finish.'
-          end
-          puts "  Database initialisation has finished, waited #{(Time.current - started).round}s."
-          {{- end }}
-
           puts 'Checking if a Ruby Tempfile can be created…'
           Tempfile.create do |_f|
             puts '  Tempfile file was created successfully.'
@@ -60,7 +38,7 @@ data:
           # Mirrors UserAgent::HttpClient#proxy_no: entries are comma separated and matched exactly.
           proxy_no = Setting.get('proxy_no').to_s.split(',').map { |entry| entry.strip.downcase }
           if proxy_no.exclude?('{{ include "zammad.elasticsearchHost" . | lower }}')
-            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . | lower }}' is missing from the 'proxy_no' setting (#{state})."
+            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . | lower }}' is missing from the 'proxy_no' setting (#{state}). Expected on a fresh install; benign if this release was installed onto an already-seeded database, or if the value was changed by an admin."
           end
           puts "  Elasticsearch host is excluded from proxy routing (#{state})."
           {{- end }}
@@ -78,8 +56,7 @@ metadata:
   annotations:
     "helm.sh/hook": test
 spec:
-  # Long enough to outlast the wait for the database initialisation above.
-  activeDeadlineSeconds: 600
+  activeDeadlineSeconds: 30
   backoffLimit: 0
   template:
     metadata:
--- a/.github/helm-force-conflicts-wrapper.sh
+++ b/.github/helm-force-conflicts-wrapper.sh
@@ -1,17 +1,23 @@
 #!/usr/bin/env bash
 #
 # Shadows the real `helm` binary on PATH during `ct install` to
-#   strip `--force-conflicts` for subcommands that don't understand it.
+#   strip flags from the subcommands that don't understand them.
 #
-# Helm 4 defaults to Server-Side Apply, which conflicts with fields (e.g.
-# .spec.nodeSets) the ECK operator takes ownership of after install, so
-# `helm upgrade` needs --force-conflicts. But chart-testing's
-# --helm-extra-args is passed to every helm subcommand it runs (install,
-# upgrade, test, uninstall), and --force-conflicts is only accepted by
-# install/upgrade/rollback - ct has no way to scope it per subcommand
-# (helm/chart-testing#540, closed as not_planned). This wrapper strips
-# --force-conflicts for any subcommand that doesn't understand it, so it can
-# stay in --helm-extra-args without breaking `helm test`/`helm uninstall`.
+# chart-testing's --helm-extra-args is passed to every helm subcommand it runs
+# (install, upgrade, test, uninstall), and ct has no way to scope it per
+# subcommand (helm/chart-testing#540, closed as not_planned). Two flags need
+# scoping, in opposite directions:
+#
+#   --force-conflicts  only install/upgrade/rollback accept it. Helm 4 defaults
+#                      to Server-Side Apply, which conflicts with fields (e.g.
+#                      .spec.nodeSets) the ECK operator takes ownership of after
+#                      install, so `helm upgrade` needs it.
+#   --logs             only `helm test` accepts it. It dumps the test pod logs,
+#                      which is the only way a failing helm test reports what it
+#                      actually saw: ct prints namespace events, never pod logs.
+#
+# This wrapper keeps each flag for the subcommands that accept it and strips it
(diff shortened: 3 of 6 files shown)


zammad/templates/configmap-init.yaml
```diff
@@ -14,6 +14,14 @@ data:
 
     if ! (bundle exec rails r 'puts User.any?' 2> /dev/null | grep -q true); then
         bundle exec rake db:migrate db:seed
+    {{- if .Values.zammadConfig.elasticsearch.initialisation }}
```
mantas, line 17: This is gated on `elasticsearch.initialisation`, but the condition it actually needs is `elasticsearch.enabled`.

With `zammadConfig.elasticsearch.enabled: false` and `initialisation: true` (the default), `zammad.elasticsearchHost` falls through to `zammadConfig.elasticsearch.host` — so an **external** ES host gets written into `proxy_no`. That is exactly the case where the proxy may be the only route to that host, and the comment two lines below ("Connections between pods of the same cluster are not external") is rendered verbatim into the script while being false for this path.

Concretely: fresh install, external ES at `es.corp.example.com`, an egress proxy configured in Zammad's `proxy` setting → `Setting.set('proxy_no', 'es.corp.example.com')` → every ES request now bypasses the proxy and fails, starting with the `SearchIndexBackend.version` call in the `elasticsearch-init` container.

```suggestion
    {{- if .Values.zammadConfig.elasticsearch.enabled }}
```

zammad/templates/configmap-init.yaml
```diff
@@ -14,6 +14,14 @@ data:
...
+
+        # Connections between pods of the same cluster are not external, so exclude Elasticsearch
+        #   from proxy routing. Only done on first install, so that admins can change the value later.
+        ELASTICSEARCH_HOST={{ include "zammad.elasticsearchHost" . }}
```
mantas: The interpolated value is unquoted in shell here, and then interpolated unescaped into a single-quoted Ruby string literal two lines down.

* `host: "es host"` (a stray space, easy to introduce in a multi-line YAML value) renders `ELASTICSEARCH_HOST=es host` — bash treats `host` as a command run with that variable in its environment, it is not found, `set -e` fires, and the init job crash-loops with an opaque `host: command not found`.
* `host: "es'.example.com"` renders `Setting.set('proxy_no', 'es'.example.com')` — broken Ruby; with a crafted value, arbitrary Ruby executed in the Rails context.

Quoting the assignment fixes the first case and is worth doing regardless; the Ruby literal still wants the value validated or escaped.

```suggestion
        ELASTICSEARCH_HOST="{{ include "zammad.elasticsearchHost" . }}"
```

mantas (commented):
Two issues from a review of this diff — the mechanism itself checks out (ES calls do route through `UserAgent`, so `proxy_no` takes effect, and `NO_PROXY_DEFAULT` keeps loopback covered), but the gate is on the wrong value and the host is interpolated unquoted.

zammad/templates/configmap-init.yaml
mantas: Seems legit? But let me know if it's just a haliucination

zammad/templates/configmap-init.yaml
mantas: Cosmetic

mantas (commented):
See some findings. My docker-fu is not enough to decide if it's legit or not.

mantas:
Whoop. My agent not only checked the code, but made a commit out of suggestions too. Sorry.

zammad/templates/configmap-init.yaml
```diff
@@ -14,6 +14,14 @@ data:
 
     if ! (bundle exec rails r 'puts User.any?' 2> /dev/null | grep -q true); then
         bundle exec rake db:migrate db:seed
+    {{- if .Values.zammadConfig.elasticsearch.enabled }}
```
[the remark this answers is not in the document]
mantas: @mgruner hm, what do you think about this suggestion?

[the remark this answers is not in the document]
mgruner: I changed it back to the original gate, as also copilot suggested. It's ok to also exclude external ES services; I believe in most cases they will also not require a proxy for accessing.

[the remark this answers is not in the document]
mgruner: I changed it back as suggested.

zammad/templates/configmap-init.yaml
```diff
@@ -14,6 +14,14 @@ data:
 
     if ! (bundle exec rails r 'puts User.any?' 2> /dev/null | grep -q true); then
         bundle exec rake db:migrate db:seed
+    {{- if .Values.zammadConfig.elasticsearch.initialisation }}
+
+        # Connections between pods of the same cluster are not external, so exclude Elasticsearch
```
fliebe92: This comment is rendered verbatim into the init script, and it is false on the `zammadConfig.elasticsearch.enabled: false` path: there the host is an external one, and the connection to it *is* external. Since the block deliberately covers that case too, the justification should not be the in-cluster one.

Suggest wording that holds for both paths:

```suggestion
        # Elasticsearch is contacted directly, not through a configured proxy. Only done on first
        #   install, so that admins can change the value later.
```

zammad/templates/configmap-init.yaml
```diff
@@ -14,6 +14,14 @@ data:
...
+        #   from proxy routing. Only done on first install, so that admins can change the value later.
+        ELASTICSEARCH_HOST="{{ include "zammad.elasticsearchHost" . }}"
+        echo "Excluding ${ELASTICSEARCH_HOST} from proxy routing..."
+        bundle exec rails r "Setting.set('proxy_no', '${ELASTICSEARCH_HOST}')"
```
fliebe92, line 23: Optional hardening. Not a security issue — the value comes from `values.yaml`, so whoever can set it can already inject init script content anyway. But a host value containing a space or a `'` fails the init job with an opaque error (`host: command not found`, or a Ruby parse error) instead of anything that points at the chart value. Passing it through the environment removes both cases:

```suggestion
        export ELASTICSEARCH_HOST="{{ include "zammad.elasticsearchHost" . }}"
        echo "Excluding ${ELASTICSEARCH_HOST} from proxy routing..."
        bundle exec rails r 'Setting.set("proxy_no", ENV.fetch("ELASTICSEARCH_HOST"))'
```

`bin/docker-entrypoint` has the same shape, so keeping it as is for parity is defensible too.

fliebe92 (commented):
One more finding that has no diff line to hang on:

**No CI coverage for this block.** `zammad/ci/full-values.yaml` leaves `zammadConfig.elasticsearch.enabled` and `initialisation` at their defaults, so only the bundled-Elasticsearch path is ever installed, and no job asserts that `proxy_no` actually ends up set. A regression here (wrong gate, wrong helper, broken interpolation) passes CI silently. Acceptable for a change this size, but worth stating explicitly rather than leaning on the `install-chart` jobs.

zammad/templates/configmap-init.yaml
fliebe92: Agree with keeping `initialisation`, and I think there is a stronger reason for it than "external ES rarely needs a proxy": it is the exact condition under which the chart writes `es_url` at all — `elasticsearch-init` is only rendered under `zammadConfig.elasticsearch.initialisation` (`configmap-init.yaml:50`, `job-init.yaml:86`). So "the chart manages Zammad's Elasticsearch config" and "the chart excludes that host from proxy routing" stay one and the same switch. Gating on `and enabled initialisation` would write `es_url` for an external host but not exclude it, which is the inconsistent combination.

It also matches the merged Docker change: `ELASTICSEARCH_ENABLED != false` seeds `proxy_no` there too, external host included. Diverging between Docker and Helm would be worse than the edge case.

The case you describe is real: fresh install, external ES only reachable through the egress proxy. It costs that admin one edit in Admin → System → Network, the value is only ever written on first install, and existing installations are untouched. That trade looks better than the inconsistency.

What does need fixing is the rendered comment, which claims the in-cluster justification on that external path — separate comment for that.

zammad/templates/configmap-init.yaml
fliebe92: Agreed, not a security issue — the value comes from `values.yaml`, from someone who can already inject init script content. The shell assignment is quoted now, which covers the `set -e` crash-loop case. The Ruby literal is still unescaped; I left a suggestion above to pass the host through the environment and use `ENV.fetch`, which closes it without any validation logic. Optional — `bin/docker-entrypoint` has the same shape.

zammad/templates/configmap-init.yaml
mgruner: Applied in 0b3b8b7, verbatim. You are right that the in-cluster justification is exactly wrong on the path the gate deliberately keeps.

zammad/templates/configmap-init.yaml
mgruner: Keeping the current shape, for parity. `bin/docker-entrypoint` has since merged unchanged (zammad/zammad!14238), still with the inline `'${ELASTICSEARCH_HOST}'` form, so switching only the chart would leave two implementations of the same thing reading differently. Since we agree it is not a security issue and the failure mode needs a quote or a space in a chart value, parity wins here. Happy to revisit both together if the entrypoint is ever hardened.

mgruner:
@fliebe92 on the CI coverage point — you were right that nothing covered this, so rather than just stating it, `0b3b8b7` adds an assertion to the helm test rake task:

```ruby
{{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
if Setting.get('proxy_no').to_s.exclude?('{{ include "zammad.elasticsearchHost" . }}')
  raise "Elasticsearch host is missing from the 'proxy_no' setting."
end
```

The `.Release.IsInstall` guard is load-bearing. `.github/ct.yaml` sets `upgrade: true`, so `ct install --upgrade` installs 18.0.5 first, then upgrades to 18.1.0 and runs the *new* chart's tests against it. That release's database was seeded by the old init job, so `User.any?` is already true when the new init job runs and the first-install branch is skipped — correctly. Without the guard the assertion would fail every upgrade leg. Since `helm test` runs the hook manifest stored at the last install/upgrade, `IsInstall` is exactly the right discriminator.

This covers the default (bundled ECK) path, which is what `ci/default-values.yaml` and `ci/full-values.yaml` install. The external-Elasticsearch path still has no install coverage — it would need a reachable external cluster — so that one remains render-checked only.

Also, for the record on Copilot's suppressed suggestion to append to `proxy_no` rather than overwrite it: not needed. `proxy_no` seeds with `state: ''` (`db/seeds/settings.rb`), so there is nothing to preserve on a first install, and `NO_PROXY_DEFAULT` (`localhost`, `127.0.0.1`, `::1`) is prepended unconditionally in `lib/user_agent/http_client.rb:78` regardless of the setting's value. Overwriting loses nothing, and it matches the merged entrypoint.

zammad/templates/tests/run-tests.yaml
```diff
@@ -30,6 +30,14 @@ data:
           puts 'Storage volume not enabled, not testing it…'
           {{- end }}
 
+          {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
+          puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
+          if Setting.get('proxy_no').to_s.exclude?('{{ include "zammad.elasticsearchHost" . }}')
+            raise "Elasticsearch host is missing from the 'proxy_no' setting."
```
fliebe92: The `IsInstall` guard is the right discriminator for the ct upgrade leg — verified that `helm template --is-upgrade` drops the block, so the upgrade leg cannot fail on the skipped first-install branch.

It does diverge from the init script's real condition though, which is `User.any?`, not "this release is an install". Where the two differ, `helm test` fails on a healthy installation: `helm install` onto an already-seeded database (external PostgreSQL, a retained PVC after `uninstall` + `install`, a docker-compose migration) has `IsInstall` true, correctly skips the seeding, and then raises here. Same for an admin who edited `proxy_no` on purpose — the external-Elasticsearch-behind-a-proxy case from the thread above — and runs `helm test` before any upgrade.

Not worth giving up the CI gate over. Suggest keeping the raise and naming the benign causes in the message, so whoever hits it is not chasing a phantom:

```suggestion
            raise "Elasticsearch host is missing from the 'proxy_no' setting. Expected on a fresh install; benign if this release was installed onto an already-seeded database, or if the value was changed by an admin."
```

Separately, `exclude?` on the line above is a substring check, while `UserAgent` matches exact host tokens or `*.`-suffixes (`lib/user_agent/http_client.rb:34-37`). `split(',').map(&:strip).exclude?(host)` would test what actually governs the routing.

mgruner: Both applied in a2c5d7c.

The exact-match point is the more important of the two — confirmed against `proxy_no` in `lib/user_agent/http_client.rb:76-80`, which splits on commas and strips before `url_use_proxy_single?` compares whole entries. A substring check would have passed on a value that does not actually govern routing. Now:

```ruby
# Mirrors UserAgent::HttpClient#proxy_no: entries are comma separated and matched exactly.
proxy_no = Setting.get('proxy_no').to_s.split(',').map { |entry| entry.strip.downcase }
if proxy_no.exclude?('{{ include "zammad.elasticsearchHost" . | lower }}')
```

I added `| lower` on the template side to match the `downcase` the runtime applies, so an external host written with capitals in `values.yaml` does not fail the assertion.

The raise message is yours, verbatim. You are right that `IsInstall` is not the init script's real condition — I picked it because it is the only signal the test pod has, and `User.any?` is unavoidably true by the time `helm test` runs. Naming the benign causes is the right trade.

One deviation worth flagging: I used an explicit block parameter rather than `it`, even though `http_client.rb` uses `it`, because this rake task runs inside whatever image `appVersion` points at and `it` needs Ruby 3.4. Not worth the coupling for one block.

Verified: `ruby -c` on the rendered rake task is clean, the assertion renders for bundled and external Elasticsearch (lowercased), and is absent for `initialisation: false` and under `--is-upgrade`.

zammad/templates/configmap-init.yaml
```diff
@@ -14,6 +14,14 @@ data:
 
     if ! (bundle exec rails r 'puts User.any?' 2> /dev/null | grep -q true); then
         bundle exec rake db:migrate db:seed
+    {{- if .Values.zammadConfig.elasticsearch.initialisation }}
+
+        # Elasticsearch is contacted directly, not through a configured proxy. Only done on first
+        #   install, so that admins can change the value later.
+        ELASTICSEARCH_HOST="{{ include "zammad.elasticsearchHost" . }}"
+        echo "Excluding ${ELASTICSEARCH_HOST} from proxy routing..."
+        bundle exec rails r "Setting.set('proxy_no', '${ELASTICSEARCH_HOST}')"
```
fliebe92, line 23: The lowercase handling and the token split match `UserAgent::HttpClient` exactly — verified the rendered task with `ruby -c`, and with `host: ES-Corp.Example.COM` it renders `es-corp.example.com`, which is what `uri.hostname.downcase` will be compared against. That part is right.

The assertion is also already earning its place: it is why CI is red, and the failure is real. `install-chart (v1.36.4, full)` fails the fresh-install leg with `Job Failed. failed: 1/1`, and the event dump for that release shows why:

```
95s  Warning  BackOff  pod/zammad-mn6zo1qh1e-init-1-8bh2r  spec.initContainers{postgresql-init}
73s  Normal   Pulled   pod/zammad-mn6zo1qh1e-init-1-8bh2r  spec.initContainers{postgresql-init}
```

`postgresql-init` failed once and restarted. If the first attempt got far enough into `db:migrate db:seed` to create users but died before the last line, the retry sees `User.any? == true`, takes the `else` branch, and `proxy_no` is never written — for the lifetime of that installation. The same BackOff shows up on the `default` leg of that run, so this is the normal behaviour of these kind clusters rather than a fluke. Note the other legs are `cancelled` by fail-fast mid-leg, so they do not tell us they would have passed; the last genuinely green run is `20786e1`, before the assertion existed.

So the write is not retry-safe, and `bin/docker-entrypoint` has the same exposure (restart of `zammad-init` after seeding) — it was just never observable there. Suggest making the write idempotent and lifting it out of the first-install branch:

```bash
bundle exec rails r "Setting.set('proxy_no', '${ELASTICSEARCH_HOST}') if Setting.get('proxy_no').blank?"
```

That survives init retries and upgrades, and still never overwrites an admin's value — the only behaviour change is that deliberately *emptying* the field gets re-seeded on the next init run. It does diverge from the entrypoint until `zammad!14238` gets the same guard.

It cannot move before `db:seed`, since the `proxy_no` row does not exist yet and `Setting.set` would raise.

The alternative is dropping the assertion back to a `puts` warning, but then the chart silently ships installations where the setting is missing, which is exactly what just happened here.

mgruner:
Root cause of the failing assertion, with evidence from run 34576918975 — worth recording because it is a property of the chart, not of this change.

**`helm install --wait` returns before the init job completes.** The init job is deliberately not a Helm hook (`job-init.yaml:10-12`, the `--wait` deadlock), and `--wait` does not wait for Jobs at all — `--wait-for-jobs` is a separate flag, in Helm 3 as well as Helm 4. (Corrected: this comment originally called it Helm 4 behaviour, which was wrong — thanks @fliebe92 for the catch.) So `helm test` starts while `postgresql-init` is still running:

| time | event |
| --- | --- |
| 08:03:32 | `helm install … --wait` starts |
| 08:04:29 | `postgresql-init` begins migrating a fresh database |
| 08:04:41 | still seeding — `Setting.set('storage_provider', "S3")`, after the seed-file loop, before locale/translation install |
| 08:05:02 | **`helm install --wait` returns, `helm test` starts** |
| 08:05:15 | assertion fails with `proxy_no=""` |

The other release in the same run confirms it independently: at 08:05:17 its init pod was still inside `elasticsearch-init` running `searchindex:rebuild`, although its `helm test` had already run and passed.

The seeding itself was healthy — the assertion reported `users=2, overviews=10, schedulers=29`. Nothing was wrong with the `proxy_no` write; the test simply looked too early.

Fixed in 9ba5903 by waiting for the initialisation rather than racing it, once for the whole suite so later database dependent checks inherit it. `elasticsearch-init` sets `es_url` and only runs after `postgresql-init` succeeded, so a present `es_url` is a usable completion signal that needs no API access from the test pod. `activeDeadlineSeconds` on the test job goes from 30 to 600 to outlast the wait.

**Beyond this PR:** the same race means `helm install --wait` reports a release ready while migrations and seeding are still in progress. That affects `ct install`, ArgoCD syncs and anything scripting `--wait` — not something this PR introduces or should fix, but it is now documented here rather than lost.

zammad/templates/tests/run-tests.yaml
```diff
@@ -13,6 +13,27 @@ data:
         desc 'Runs a set of Helm tests'
         task test: :environment do |_task, args|
 
+          {{- if .Values.zammadConfig.elasticsearch.initialisation }}
```
fliebe92: The wait is gated on `initialisation` alone, while its only consumer below also requires `.Release.IsInstall`. On an upgrade the loop is a no-op whenever `es_url` is set — but when it legitimately is not (ES was down at install time, or an admin cleared the value), `helm test` spends 8 minutes and then fails with `Timed out waiting for the database initialisation to finish.`, which is a misleading report of an unrelated state. Suggest gating it the same way, so the wait exists exactly where its result is used:

```suggestion
          {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
```

zammad/templates/tests/run-tests.yaml
```diff
@@ -13,6 +13,27 @@ data:
...
+          # `helm install --wait` returns before the init job completes, because the job is
+          #   deliberately not a Helm hook (see job-init.yaml), so this test can start while
+          #   postgresql-init is still migrating and seeding. elasticsearch-init sets es_url and
+          #   only runs once postgresql-init succeeded, so a present es_url means the database
+          #   initialisation is done and any check below can rely on its result.
+          # Setting.get caches for ZAMMAD_SETTING_TTL (15s), so force a reload on every attempt.
+          puts 'Waiting for the database initialisation to finish…'
+          deadline = 8.minutes.from_now
```
fliebe92: 8 minutes of waiting plus Rails boot lands at roughly 8.5 minutes against the job's `activeDeadlineSeconds: 600`, so the margin is ~1.5 minutes — and if the wait ever does time out, Kubernetes kills the pod with `DeadlineExceeded` and the explicit `Timed out waiting…` message never reaches the log. Either shorten this to `5.minutes.from_now`, or raise `activeDeadlineSeconds` to 900 so it matches ct's `helm test --timeout 900s` and the Ruby message always wins.

zammad/templates/tests/run-tests.yaml
```diff
@@ -30,6 +51,19 @@ data:
...
 
+          {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
+          puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
+          # Seed counts tell apart "db:seed never finished" from "the write itself did not happen":
+          #   users appear with the 2nd seed file, overviews and schedulers near the end.
+          state = "proxy_no=#{Setting.get('proxy_no').inspect}, system_init_done=#{Setting.get('system_init_done').inspect}, users=#{User.count}, overviews=#{Overview.count}, schedulers=#{Scheduler.count}"
+          # Mirrors UserAgent::HttpClient#proxy_no: entries are comma separated and matched exactly.
+          proxy_no = Setting.get('proxy_no').to_s.split(',').map { |entry| entry.strip.downcase }
+          if proxy_no.exclude?('{{ include "zammad.elasticsearchHost" . | lower }}')
+            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . | lower }}' is missing from the 'proxy_no' setting (#{state})."
```
fliebe92: The benign-cause hint dropped out here. A fresh `helm install` onto an already-seeded database — existing external PostgreSQL, or a docker-compose migration — has `es_url` set already, so the wait above returns immediately and this raises on a perfectly healthy installation. The `state` dump identifies it for us, but it does not tell an admin that this case is expected. Suggest keeping both:

```suggestion
            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . | lower }}' is missing from the 'proxy_no' setting (#{state}). Expected on a fresh install; benign if this release was installed onto an already-seeded database, or if the value was changed by an admin."
```

zammad/templates/configmap-init.yaml
fliebe92: You are right and I was wrong about the cause: this is the `--wait` race, not the init retry. `job-init.yaml:10-12` explains why the job is not a hook, and ct runs `helm install --wait` without `--wait-for-jobs`, so Helm never waits for it; `railsserver` gates only on pending migrations, so the release reports ready while `db:seed` is still running. The `BackOff` I quoted was concurrent noise.

The retry exposure is still there though, just no longer the thing breaking CI: if `postgresql-init` dies after `db:seed` created users, the retry takes the `else` branch and the write is lost for the lifetime of that installation. The new `state` dump is enough to recognise it if it ever shows up (`users>0` with `proxy_no=""`), so I am fine leaving the `blank?` guard out and keeping parity with the entrypoint. Worth remembering as the explanation if this assertion ever fails in the wild.

zammad/templates/tests/run-tests.yaml
mgruner: Applied, but by removing the wait rather than gating it — your `--wait-for-jobs` point below made the whole block unnecessary. `ct install` now passes it, so the init job has finished before `helm test` starts. That also removes the `activeDeadlineSeconds` margin problem you raised in the next comment.

zammad/templates/tests/run-tests.yaml
mgruner: Resolved by removing the wait entirely (see above), so `activeDeadlineSeconds` goes back to 30 and there is no Ruby-message-versus-`DeadlineExceeded` race left. Your reading of the margin was right, though: 8 minutes plus a Rails boot against a 600s deadline left almost nothing, and `DeadlineExceeded` would have won.

zammad/templates/tests/run-tests.yaml
mgruner: Applied verbatim in dbb000e. You are right that it still reaches a healthy installation — a fresh `helm install` onto an existing database has `proxy_no` empty and nothing in the state dump says that is expected. Dropping the hint when I added `state` was an oversight, not a decision.

zammad/templates/configmap-init.yaml
mgruner: `--wait-for-jobs` is the key correction, thank you — and it means my earlier comment on this PR was wrong to call it Helm 4 behaviour. `--wait` has never waited for Jobs; `--wait-for-jobs` is a separate flag and exists in Helm 3 too. I have corrected that comment.

So the fix moved out of the chart: `ct install` now passes `--wait-for-jobs` (scoped by the existing wrapper, since `helm test` and `helm uninstall` reject it), and the in-test wait is gone. ct's install test also stops reporting success on half-initialised releases, which it was doing for every scenario before.

Agreed on leaving the `blank?` guard out and keeping parity with the entrypoint. For the record on the retry exposure: `db/seeds.rb` loads `user_nr_1` second, so `User.any?` flips true within seconds of `db:seed` starting — the unprotected window is most of the seed, not just the tail. If the assertion ever fails with `users>0` and `proxy_no=""`, that is the signature.

The README now documents `--wait-for-jobs` for anyone scripting installs or using ArgoCD.

mgruner:
Measured the `--wait-for-jobs` approach and reverted it — it was correct but far too expensive, so the wait is back inside the test, now gated on `.Release.IsInstall` as @fliebe92 suggested.

Per-leg phase totals, `install-chart (v1.37.0, full)`:

| approach | install | upgrade | tests | total |
| --- | --- | --- | --- | --- |
| no wait (`20786e1`) | 109/127/126s | 19/2s | 13–15s | **454s** |
| wait in the test | 96/129/76s | 23/1s | 15–51s | **498s** |
| `--wait-for-jobs` | 195/192/195s | 91/83s | 15–17s | **834s** |

`--wait-for-jobs` makes *every* `helm install` and `helm upgrade` block on a complete init job cycle — three Rails boots plus migrate/seed plus Elasticsearch index work — including the previous-revision install and both upgrades, which never needed to wait for this assertion. Six and a half minutes per leg for a gate that costs 44 seconds in the test. It does leave a real gap: `ct install` can still report success on a half-initialised release. Documented rather than fixed, since closing it costs that much.

The measured waits also settle two earlier open questions. They are **25–86s**, so the cost is genuine initialisation time rather than polling overhead; and the post-upgrade invocations really were waiting, because without `--wait-for-jobs` the previous revision's init job was often still running when ct upgraded, so `es_url` was not yet set.

The 86s matters for a second reason: it is past the 80s ceiling (`failureThreshold: 20` × `periodSeconds: 4`) of the railsserver startup probe that the test container was inheriting through `containerSpec`. The rake pod never listens on that port, so the probe could only ever kill it. That leg passed only because the probes had already been removed — hence that change staying in.

`--wait-for-jobs` is still the right advice for anyone scripting installs or using ArgoCD, so it stays documented in the README.

fliebe92 (dismissed):

zammad/templates/tests/run-tests.yaml
mgruner: Reworked in 4baa619, so this thread's suggestion is resolved differently than my last reply said. The wait is back, but it no longer keys on `es_url`. `postgresql-init` now writes a revision-scoped flag to the Rails cache once it finishes, and the test waits for that:

```ruby
Rails.cache.read('helm-postgresql-init-done-{{ .Release.Revision }}')
```

Your underlying point still drove it: the wait should exist only where it can succeed. It is now gated on `zammadConfig.initJob.enabled`, so with no init job to wait for it is compiled out entirely — the case `.Release.IsInstall` would not have covered. And because the key carries `.Release.Revision`, a previous revision's flag cannot satisfy it on an upgrade, which was the other thing the `IsInstall` gate was protecting against.

zammad/templates/tests/run-tests.yaml
mgruner: Still 5 minutes, and `activeDeadlineSeconds` stays at 600, so there is ~5 minutes of headroom for the Rails boot and the remaining checks rather than the ~1.5 you flagged. The explicit `Timed out waiting…` message wins over `DeadlineExceeded` in every realistic case now.

zammad/templates/tests/run-tests.yaml
mgruner: Being straight with you: the hint is not in 4baa619. The message was shortened to the host plus the actual value:

```
Elasticsearch host 'x-es-http' is missing from the 'proxy_no' setting (#{proxy_no}).
```

The case you described is still reachable — a fresh `helm install` against an already-seeded database (external PostgreSQL, a docker-compose migration) has `IsInstall` true, takes the `else` branch, and raises here on a healthy installation. So the reasoning behind your suggestion still stands; what changed is the judgement that printing the observed value carries most of the diagnostic weight. Say the word if you would rather have the sentence back.

zammad/templates/configmap-init.yaml
mgruner: Final state on this, since the approach changed twice since your comment.

`--wait-for-jobs` was implemented and then reverted: correct, but it makes *every* install and upgrade block on a full init cycle, including phases with no stake in this assertion. Measured on the `v1.37.0/full` leg, per-leg phase totals were 454s with no wait, 498s waiting inside the test, and 834s with `--wait-for-jobs` — six and a half minutes for a gate that costs well under one. Numbers in https://github.com/zammad/zammad-helm/pull/470#issuecomment-5632606782. It stays documented in the README, since it remains the right advice for anyone scripting installs or using ArgoCD.

4baa619 replaces the `es_url` heuristic with an explicit signal: `postgresql-init` writes `helm-postgresql-init-done-<revision>` to the Rails cache once it completes, and the test waits for that. It is a more honest completion marker than `es_url` (which is set midway through `elasticsearch-init`), it releases before the search index rebuild rather than after, and the revision in the key means an older flag cannot satisfy a newer test. Cost is one extra `rails r` per init run, which delays only the job — service pods gate on `RAILS_CHECK_PENDING_MIGRATIONS`, satisfied by `db:migrate` earlier in the same container.

One property worth recording: `ActiveSupport::Cache::MemCacheStore` wraps writes in `rescue_error_with` (`mem_cache_store.rb:278`), so a memcached outage cannot fail the init job under `set -e` — the flag simply never appears and `helm test` times out. Test-only consequence, and the trade we are accepting for not adding a settings row.

Your note about the retry exposure still holds and is still not addressed, by agreement: if `postgresql-init` dies after `db:seed` created users, the retry takes the `else` branch and the write is lost for that installation's lifetime.

zammad/templates/tests/run-tests.yaml
mgruner: Heads-up that 4baa619 replaced this with a plain equality check:

```ruby
proxy_no = Setting.get('proxy_no')
if proxy_no != '{{ include "zammad.elasticsearchHost" . }}'
```

Your point about `UserAgent::HttpClient` matching whole comma-separated entries is still correct — this is simply stricter than the runtime semantics, and deliberately so: the assertion only runs on a fresh install (`.Release.IsInstall`), where the chart has just written exactly that one host, so anything else is a finding. It does mean a value with the host *plus* another entry now fails, where the previous version passed. The `| lower` normalisation is gone too; both sides render the same helper, so they agree, but it no longer mirrors the runtime's `downcase`.

The seed-count dump went with it. It existed to tell an aborted `db:seed` from a missing write, and the cause it was diagnosing turned out to be the `--wait` race instead.

mgruner:
Root cause of the intermittent `proxy_no=""` failures, fixed in e1479ba. It was a stale settings cache in the test process — the init job was correct the whole time.

From the failing test pod's own log:

```
12:22:28.28  Using memcached as Rails cache store          ← boot
12:22:29.78  ActionCable is configured …                   ← first Setting read, @@current populated
12:22:30.07  Waiting for the database initialisation to finish…
12:22:40.07    Database initialisation has finished, waited 10s.
12:22:40.07  Elasticsearch host '…' is missing from the 'proxy_no' setting ().
```

And from the same release's init job, which ran to completion:

```
12:22:24.96  Excluding zammad-rt0c5lpvaq-es-es-http from proxy routing...
12:22:38.32  postgresql init complete :)
```

`Setting.load` populated `@@current` at 12:22:29.8, before the init job's `Setting.set('proxy_no', …)` landed. `cache_valid?` short-circuits while `@@lookup_at` is newer than `@@lookup_timeout` (`ZAMMAD_SETTING_TTL`, 15s by default), so the read at 12:22:40.1 — 10.3s later — was served from the in-process cache and never re-queried. The database already held the right value. One `Setting.reload` before the check fixes it.

Two things worth recording:

**The `es_url` version never hit this by accident.** Its wait loop called `Setting.reload` on every attempt to see `es_url` appear, which incidentally left the later `proxy_no` read fresh. Replacing the loop with a `Rails.cache` flag removed the wait's dependency on settings, and with it the side effect that was keeping the assertion honest.

**Two independent bugs produced an identical symptom.** `proxy_no=""` on a healthy install was first the `--wait` race (`helm install --wait` does not wait for Jobs) and then this. Fixing the first exposed the second, which is why this took several rounds.

@fliebe92 — your retry-safety concern is *not* what was failing here: the init job demonstrably took the first-install branch and wrote the value. The exposure you described is still real and still unaddressed by agreement, and the `Excluding …` line in the init log is the way to tell the two apart if it ever does show up.

zammad/templates/tests/run-tests.yaml
```diff
@@ -13,6 +13,20 @@ data:
...
+          (started, deadline)  = Time.current, 5.minutes.from_now
+          while Time.current < deadline
+            break if Rails.cache.read('helm-postgresql-init-done-{{ .Release.Revision }}')
+            sleep 5
+          end
+          if !Rails.cache.read('helm-postgresql-init-done-{{ .Release.Revision }}')
+            raise 'Timed out waiting for the database initialisation to finish.'
+          end
+          puts "  Database initialisation has finished, waited #{(Time.current - started).round}s."
+          {{- end }}
```
fliebe92: `Rails.cache` is not a durable signal, and this block runs on every `helm test` (it is gated on `initJob.enabled`, not on the install). The marker expires after an hour, `Rails.cache.clear` runs during ordinary operation — `app/models/user.rb:1126` on user deletion, `app/models/object_manager/attribute.rb:797` on attribute migration — and memcached restarts or evicts. So `helm test` on a release older than an hour blocks 5 minutes and then reports `Timed out waiting for the database initialisation to finish.` on a fully initialised installation.

`.Release.IsInstall` would not fix it: it stays true in the stored manifest until the next upgrade, so a test run days after the install still takes this path.

Cheapest fix is to not raise when the database is demonstrably initialised anyway — on timeout, fall back to "no pending migrations and `User.any?`" and only raise if that fails too. The marker then still gives the fast, precise signal right after an install, without turning into a false failure later.

zammad/templates/configmap-init.yaml
```diff
@@ -22,6 +30,9 @@ data:
         bundle exec rails r "Locale.sync; Translation.sync"
     fi
 
+    # Also store a value in the Rails cache that can be used to detect job completion.
+    bundle exec rails r "Rails.cache.write('helm-postgresql-init-done-{{ .Release.Revision }}', true, { expires_in: 1.hour })"
```
fliebe92: This is a third unconditional Rails boot in `postgresql-init`, paid on every install and every upgrade — and by your own measurements a boot is a noticeable share of the init cost.

The init job's main container already runs `bundle exec rails r "true"` in `postgresql-init-post` purely as a readiness no-op. Writing the marker there instead is free, and it strengthens the signal at the same time: `postgresql-init-post` runs after *all* init containers, so the key would mean "the whole init job finished" rather than "postgresql-init finished".

zammad/templates/tests/run-tests.yaml
```diff
@@ -30,6 +44,16 @@ data:
           puts 'Storage volume not enabled, not testing it…'
           {{- end }}
 
+          {{- if and .Values.zammadConfig.elasticsearch.initialisation .Release.IsInstall }}
+          puts 'Checking if the Elasticsearch host is excluded from proxy routing…'
+          Setting.reload # Settings cache might be stale, so refresh.
+          proxy_no = Setting.get('proxy_no')
+          if proxy_no != '{{ include "zammad.elasticsearchHost" . }}'
+            raise "Elasticsearch host '{{ include "zammad.elasticsearchHost" . }}' is missing from the 'proxy_no' setting (#{proxy_no})."
```
fliebe92: Exact equality is stricter than the routing it verifies. `UserAgent::HttpClient` splits `proxy_no` on commas and matches per entry (`http_client.rb:34-37`, `:78`), so any additional legitimate entry fails here while routing is correct — an autowizard payload that sets `proxy_no`, or an admin editing the field between the install and the test run. A `split(',').map(&:strip)` membership check keeps the assertion honest at the same cost.

Minor, related: the message now prints the actual value, which is good, but no longer says that a mismatch can be legitimate (installed onto an already-seeded database, or the value was changed on purpose).

zammad/templates/tests/run-tests.yaml
mgruner: AFAIK Helm test is not expected to be run outside of CI, that's why the current implementation and timing seems sufficient. Are there other use cases?

zammad/templates/configmap-init.yaml
mgruner: I decided against this, because postgresql-init-post runs after the ES jobs and would significantly slow down CI.
In postgresql-init it does cost a few seconds, but that does not delay other production pods who just wait for migrations to be ready.

zammad/templates/tests/run-tests.yaml
mgruner: Again, the expectation is/was that this check only runs in CI, and can therefore be simplified to the expected initial value.

zammad/templates/tests/run-tests.yaml
fliebe92: For the automated paths you are right — Flux runs `helm test` straight after install/upgrade via `HelmRelease.spec.test.enable`, so the marker is always fresh there, and ArgoCD does not run tests at all.

The case I would not dismiss is inside this same rake task: the Tempfile and storage-volume checks are exactly what someone runs against an existing release to answer "can Zammad still write to its storage volume?". The wait now sits in front of them, so an hour after install that diagnostic spends 5 minutes and then fails with a message about database initialisation, which has nothing to do with what was being checked. The assertion below needs the wait; those checks do not.

Two ways out, both cheap. Either move the wait inside the `initialisation`/`IsInstall` block so it only guards the check that needs it, or keep it where it is and fall back on timeout to "no pending migrations and `User.any?`", raising only if that fails too.

If the intent is that these tests only ever run right after an install, that is a fine decision — but then it is worth one line in the README, because nothing in the chart says so today and `helm test` is a standard command against any release.

zammad/templates/configmap-init.yaml
fliebe92: Fair — I had not accounted for the wait then covering the Elasticsearch init as well. Dropping that suggestion.

One cheaper leftover, if you want it: the `else` branch already boots Rails for `Locale.sync; Translation.sync`, so appending `Rails.cache.write(...)` to that command costs nothing on upgrades, which is the path production users pay on every release. Only the first-install path would still need the standalone call. Entirely optional — it trades a few template lines for one Rails boot per upgrade.

zammad/templates/tests/run-tests.yaml
fliebe92: Agreed as a decision — and it is the same premise as the wait thread, so both of my points collapse into one question: is "these tests only run in CI, right after an install" written down anywhere? Right now `helm test` is a standard command against any release, the chart ships `templates/tests/` without any note, and the Tempfile/storage checks in the same task look like general diagnostics.

One line stating the assumption — in the README next to the install instructions, or as a comment at the top of the rake task — makes both the simplified comparison and the cache-based wait obviously correct rather than surprising. With that, I am happy to close both threads.

zammad/templates/tests/run-tests.yaml
mgruner: I researched a bit - it can be a use case to run 'helm test' in production. Then a config seeding test like this one is inappropriate - I dropped it again. Please have another look.

zammad/templates/tests/run-tests.yaml
```diff
@@ -43,7 +43,9 @@ metadata:
   annotations:
     "helm.sh/hook": test
 spec:
-  activeDeadlineSeconds: 30
+  # Outlasts the wait for the database initialisation above, so that its own timeout message
+  #   reaches the log instead of Kubernetes killing the pod with DeadlineExceeded first.
+  activeDeadlineSeconds: 600
```
fliebe92: This deadline outlived its reason: the comment points at "the wait for the database initialisation above", which `d5b3f78` removed. The task now boots Rails and runs three file checks — 11-17s in the CI logs.

Either back to `30` as it was, or keep a larger value with a rationale that still holds. Worth weighing for the second option: with the probes now omitted, nothing else caps a hung test pod, so `600` means a hang takes ten minutes to surface instead of being killed at 80s.

fliebe92:
The description no longer matches the diff after `d5b3f78`. Four paragraphs still explain the helm test assertion, its `.Release.IsInstall` guard, the `helm-postgresql-init-done-<revision>` cache marker and the wait built on it, and the `--wait-for-jobs` experiment — all of which are out of the PR now. The parts of that story that survived are the README note and the probe fix in `run-tests.yaml`.

Worth rewriting before merge, since this text becomes the merge commit: what lands is the `proxy_no` seeding itself, the note that `--wait` does not wait for Jobs, and the probe removal as a drive-by fix. The reason the assertion was dropped is also worth one line — `helm test` can run against a long-lived production release, where a first-install-only check cannot be correct — so the next person does not re-add it.

Also: the seeding is render-checked only again, with no CI assertion behind it. That is a reasonable trade given the above, but it is a different coverage claim than the one the body currently makes.

zammad/templates/tests/run-tests.yaml
mgruner: Good point, I dropped the comment.

mgruner:
> The description no longer matches the diff after `d5b3f78`. Four paragraphs still explain the helm test assertion, its `.Release.IsInstall` guard, the `helm-postgresql-init-done-<revision>` cache marker and the wait built on it, and the `--wait-for-jobs` experiment — all of which are out of the PR now. The parts of that story that survived are the README note and the probe fix in `run-tests.yaml`.
> 
> Worth rewriting before merge, since this text becomes the merge commit: what lands is the `proxy_no` seeding itself, the note that `--wait` does not wait for Jobs, and the probe removal as a drive-by fix. The reason the assertion was dropped is also worth one line — `helm test` can run against a long-lived production release, where a first-install-only check cannot be correct — so the next person does not re-add it.
> 
> Also: the seeding is render-checked only again, with no CI assertion behind it. That is a reasonable trade given the above, but it is a different coverage claim than the one the body currently makes.

Description updated.

zammad/templates/tests/run-tests.yaml
fliebe92: The comment is gone but the value stayed, so the diff now carries a bare `30 → 600` with nothing saying why. It is a defensible number — the probes are omitted now, so nothing else caps a hung test pod, and Rails boot alone is 11-17s in the CI logs, which made `30` thin — but that reasoning only exists in this thread. Either a one-liner in the template, or a sentence in the description, so the next person does not have to dig it out of a closed PR.

fliebe92:
Body reads accurately for the seeding now. One gap left: it covers only the seeding, while the PR also ships the probe `omit` in `run-tests.yaml`, the `activeDeadlineSeconds` bump, and the README note that `--wait` does not wait for Jobs. Those are the surviving pieces of the dropped assertion work and they are worth a sentence each, otherwise the merge commit explains none of them.

zammad/templates/tests/run-tests.yaml
mgruner: Closing this one out: the assertion has been dropped altogether (d5b3f78), so the equality check this thread was about no longer exists. Your original point stood throughout — `UserAgent::HttpClient` matches whole comma-separated entries — it just has nothing left to apply to.

zammad/templates/tests/run-tests.yaml
mgruner: The wait is gone with the assertion (d5b3f78), so this is moot. For the record, your instinct was right twice over: the wait did need gating, and in its final form it was gated on `zammadConfig.initJob.enabled` — but it also turned out that neither gate was the real problem. The assertion was failing on a stale settings cache, not on anything the wait could fix.

zammad/templates/tests/run-tests.yaml
mgruner: The wait is gone (d5b3f78), so there is no timeout message left to lose to `DeadlineExceeded`. `activeDeadlineSeconds` stays at 600 deliberately — the tests run in ~15s, so it is pure headroom for a slow Rails boot rather than a tight margin. Happy to put it back to 30 if you would rather not carry the change.

zammad/templates/tests/run-tests.yaml
mgruner: Moot now — the assertion was dropped in d5b3f78, so there is no raise message to carry the hint. The scenario you described (a fresh `helm install` onto an already-seeded database) would have kept tripping it, which is part of why the check was not worth its weight.

zammad/templates/configmap-init.yaml
mgruner: Final state on this thread. The test-side machinery is gone (d5b3f78): no wait, no completion flag, no `proxy_no` assertion. `configmap-init.yaml` carries only the seeding itself, exactly as you and @mgruner agreed — written on first install, no `blank?` guard, parity with the entrypoint.

What the dropped assertion produced is still worth keeping in view:

- `helm install --wait` does not wait for Jobs; `--wait-for-jobs` is a separate flag. The README documents it now, since it affects anyone scripting installs or using ArgoCD. `--wait-for-jobs` in `ct install` was tried and reverted: correct, but ~6.5 minutes per leg against a few seconds for the alternative.
- The retry exposure you raised is unchanged and deliberately not fixed. If `postgresql-init` dies after `db:seed` created users, the retry takes the `else` branch and the write is lost for that installation. `db/seeds.rb` loads `user_nr_1` second, so the unprotected window is most of the seed, not just its tail.

Worth noting that it was never this exposure that made CI red — the init job demonstrably took the first-install branch and wrote the value, and `Excluding <host> from proxy routing...` in the init log is the line that tells the two apart if it ever does bite.

mgruner:
@fliebe92 on your CI-coverage point, since it is the one thing this PR ends up not doing: the assertion that answered it has been dropped (d5b3f78), and the `proxy_no` block is render-checked only, as the external-Elasticsearch path already was. Reasoning is in the updated description.

Short version: it never caught a defect in the change. Every failure it produced came from the harness around it — `helm test` racing the init job because `--wait` does not cover Jobs, then a stale in-process settings cache once that was fixed. Both are real findings and both are written up above, but the check itself was costing more review rounds than the one-line shell change it guarded.

You were right that the gap exists, and it still does. It is now stated in the description rather than left implicit, which was your original ask.

fliebe92 (approved):

PR #473, opened 2026-09-17
Maintenance: clean up obsolete bitnami references after major release

mgruner:

- Clean up remaining bitnami references after the release. This was not possible before due to upgrade testing in CI.


fliebe92 (approved):
Verified the precondition for this cleanup: `main` now carries chart 19.0.0 (released as `zammad-19.0.0`), so the `ct` upgrade test no longer installs an 18.x chart with the `bitnami/minio` dependency, and the `bitnami` chart-repo entry is obsolete. The removed release-workflow step only added `bitnami`; the remaining `elastic` and `rustfs` dependencies were never added there and the 19.0.0 release packaged fine without them, so dropping the step is safe. The only remaining `bitnami` mentions are the historic migration and upgrade notes in `zammad/README.md`, which should stay. LGTM.

